PR #4, opened 2023-01-18
[Bugfix] fix initialization

szhengac:
## Description ##
1. Remove redundant dist group creation
2. Remove group destroy which cause nccl error
3. Add user defined init func.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/examples/gpt/deepspeed_hf.py
+++ b/examples/gpt/deepspeed_hf.py
@@ -142,7 +142,7 @@ def train(args):
             target="deepspeed",
             config=ds_config_dict,
             loss_fn=loss_fn,
-            param_init_fn=model._init_weights,
+            init_weights=model._init_weights,
         )
     else:
         if batch_size is not None and micro_batch_size is None:
@@ -159,7 +159,7 @@ def train(args):
             topology=topology,
             target="deepspeed",
             config=ds_config_dict,
-            param_init_fn=model._init_weights,
+            init_weights=model._init_weights,
         )
         model = model.to(device)
     report_memory(msg="After building model")
--- a/examples/gpt/deepspeed_hf.py
+++ b/examples/gpt/deepspeed_hf.py
@@ -173,8 +173,6 @@ def train(args):
 
     loader = RepeatingLoader(train_loader)
 
-    model.load_checkpoint("checkpoint", "global_step40")
-
     num_iters = args.iter_nums
     if enable_pipeline:
         data_iter = iter(loader)
@@ -182,8 +180,6 @@ def train(args):
             model.train_batch(data_iter=data_iter)
     else:
         train_with_torch(model, loader, steps=num_iters)
-
-    model.save_checkpoint("checkpoint")
 
 
 if __name__ == "__main__":
--- a/examples/gpt/deepspeed_hf.py
+++ b/examples/gpt/deepspeed_hf.py
@@ -142,6 +142,7 @@ def train(args):
             target="deepspeed",
             config=ds_config_dict,
             loss_fn=loss_fn,
+            param_init_fn=model._init_weights
         )
     else:
         if batch_size is not None and micro_batch_size is None:
@@ -158,6 +159,7 @@ def train(args):
             topology=topology,
             target="deepspeed",
             config=ds_config_dict,
+            param_init_fn=model._init_weights
         )
         model = model.to(device)
     report_memory(msg="After building model")
--- a/examples/gpt/deepspeed_hf.py
+++ b/examples/gpt/deepspeed_hf.py
@@ -124,7 +124,7 @@ def train(args):
         batch_size = 16 if batch_size is None else batch_size
         micro_batch_size = 4 if micro_batch_size is None else micro_batch_size
         ds_config_dict = get_ds_config(
-            batch_size, micro_batch_size, True, False, "Pipeline"
+            batch_size, micro_batch_size, True, False, "Pipeline", False
         )
         loss_fct = ParallelCrossEntropy(group=group)
 
@@ -173,6 +173,8 @@ def train(args):
 
     loader = RepeatingLoader(train_loader)
 
+    model.load_checkpoint("checkpoint", "global_step40")
+
     num_iters = args.iter_nums
     if enable_pipeline:
         data_iter = iter(loader)
@@ -180,6 +182,8 @@ def train(args):
             model.train_batch(data_iter=data_iter)
     else:
         train_with_torch(model, loader, steps=num_iters)
+
+    model.save_checkpoint("checkpoint")
 
 
 if __name__ == "__main__":
--- a/examples/gpt/deepspeed_hf.py
+++ b/examples/gpt/deepspeed_hf.py
@@ -142,7 +142,7 @@ def train(args):
             target="deepspeed",
             config=ds_config_dict,
             loss_fn=loss_fn,
-            param_init_fn=model._init_weights
+            param_init_fn=model._init_weights,
         )
     else:
         if batch_size is not None and micro_batch_size is None:
@@ -159,7 +159,7 @@ def train(args):
             topology=topology,
             target="deepspeed",
             config=ds_config_dict,
-            param_init_fn=model._init_weights
+            param_init_fn=model._init_weights,
         )
         model = model.to(device)
     report_memory(msg="After building model")
--- a/slapo/autotune/tune.py
+++ b/slapo/autotune/tune.py
@@ -282,7 +282,7 @@ def run_training_script(args, tuneable_cfg):
     # Set all tunable parameters as environment variables.
     env = " ".join(f"{k}={v}" for k, v in tuneable_cfg.items())
 
-    cmd = f"DEBUG=1 {env} python3 {args.training_script} {train_script_args}"
+    cmd = f"{env} python3 {args.training_script} {train_script_args}"
     cmd += " > run_script.log 2>&1"
     logger.info(f"\tRunning command: {cmd}")
     os.system(cmd)
--- a/slapo/schedule.py
+++ b/slapo/schedule.py
@@ -798,8 +798,7 @@ def create_schedule(
 def consolidate_model(
     sch: Schedule,
     topology=None,
-    param_init_fn: Optional[Callable[[nn.Module], None]] = None,
-    init_required: bool = True,
+    param_init_fn: Optional[Callable[[nn.Module], None]] = None
 ):
     if dist.get_world_size() > sch.world_size:
         assert (
@@ -885,7 +884,7 @@ def consolidate_model(
             )
 
         # use original shape to initialize parameters
-        if global_rank == curr_stage_devices[0] and num_params > 0 and init_required:
+        if global_rank == curr_stage_devices[0] and num_params > 0:
             # only the first device in the PP group needs to initialize the weights
             if param_init_fn:
                 param_init_fn(sch.mod)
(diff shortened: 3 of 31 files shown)


comaniac (commented):
Let @chhzh123 double check.
Also please rebase onto the latest main branch so that the CI should work.

chhzh123:
Is the long initialization time mainly because of the dist group creation and destruction?

slapo/schedule.py
```diff
@@ -823,30 +826,39 @@ def consolidate_model(sch: Schedule, topology=None):
             # [[0, 2], [1, 3], [4, 6], [5, 7]]
             # >>> topo.filter_match(pipe=0)
             # [0, 1, 2, 3]
-            curr_part_idx = sch.partition_idx
-            # topology stores the global ranks
-            curr_stage_devices = topology.filter_match(pipe=curr_part_idx)
             # create dist group for broadcasting
             num_pp = topology.get_dim("pipe")
             # each group contains the devices on the same stage
-            if not "DEBUG" in os.environ:
```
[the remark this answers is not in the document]
chhzh123: DEBUG mode can speedup the autotuning process and is used in the tuning script. If DEBUG mode is removed, those scripts in the benchmark folder should also be modified.

szhengac:
> Is the long initialization time mainly because of the dist group creation and destruction?

Mainly because of group creation. Group destruction causes nccl error for gpt example

slapo/schedule.py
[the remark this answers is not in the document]
szhengac: which benchmark scripts?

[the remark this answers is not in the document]
chhzh123: tune_single_xxx.py

examples/gpt/deepspeed_hf.py
```diff
@@ -171,6 +173,8 @@ def loss_fn(outputs, labels):
 
     loader = RepeatingLoader(train_loader)
 
+    model.load_checkpoint("checkpoint", "global_step40")
```
comaniac, line 176: I suppose this is for testing checkpoint? If so, I'd suggest removing it and creating a unit test.

slapo/schedule.py
```diff
@@ -790,25 +791,33 @@ def is_module_list(module):
     return root_sch
 
 
-def consolidate_model(sch: Schedule, topology=None):
+def consolidate_model(
+    sch: Schedule,
+    topology=None,
+    param_init_fn: Optional[Callable[[nn.Module], None]] = None,
+):
+    if dist.get_world_size() > sch.world_size:
```
comaniac, line 799: The use of `dist` in this function makes `consolifate_model` require `dist` to be initialized. In other words, single device use cases will crash right after calling this function. Since now whether to call this function is by user config, we have to provide a single device fallback.

slapo/schedule.py
```diff
@@ -899,27 +913,35 @@ def consolidate_model(sch: Schedule, topology=None):
...
     return sch
 
 
-def build(sch: Schedule, topology=None, target=None, **kwargs):
+def build(
+    sch: Schedule,
+    topology=None,
+    target=None,
+    param_init_fn=None,
+    init_required: bool = True,
```
comaniac, line 934: `init_required` is a bit vague, and it is tightly coupled with `param_init_fn`. Can we name it like `init_weights: Optional[Union[bool, Callable]]`? It could be used as follows:

```python
# case 1: disable
init_weights=None or False
# case 2: enable with default init function
init_weights=True
# case 3: enable with custom init function
init_weights=model._init_weights
```

tests/test_shard.py
```diff
@@ -37,14 +38,14 @@ def sync_model_params(model):
 
 def gather_grad(model, param_path_and_gather_axis):
     world_size = dist.get_world_size()
-    rank = dist.get_rank()
+    local_rank = int(os.environ["LOCAL_RANK"])
```
comaniac, line 41: Out of curiosity, what's the different between these two?

comaniac (commented):
Mostly LGTM. I'll approve when the comments are addressed.
@chhzh123 PTAL.

examples/gpt/deepspeed_hf.py
szhengac: oh this is added accidentally. This shouldn't be included in slapo repo.

slapo/schedule.py
szhengac: I think it would be misleading for consolidation func. In consolidation func, param_init_fn should always be either None or an actual func, while init_required is used in build func to indicate if we would like to call consolidation func.

and using the example as you suggested, the init will be disabled by defaults. but I would expect users want initialization when they call build since build is mainly for the pipeline case. and if users are not familiar with the code, they may forget to enable that.

tests/test_shard.py
szhengac: local_rank means the rank on the local node. For example, rank = 8 -> local_rank = 0 when each node has 8 gpus.

slapo/schedule.py
szhengac: makes sense

slapo/schedule.py
comaniac: `I would expect users want initialization when they call build since build is mainly for the pipeline case` that's not true. We expect users to call `build` in all cases. Although currently only pipeline needs it, we don't know if there's any requirements in the future. If you prefer to have this by default, we just need `init_weight: Union[bool, Callable]=True`, so that it becomes:

```python
# case 1 (default): enable with default init function
init_weights=True
# case 2: enable with custom init function
init_weights=model._init_weights
# case 3: disable
init_weights=False
```

tests/test_shard.py
comaniac: Oh nvm. I missed that `get_rank()` returns the global rank.

slapo/schedule.py
szhengac: ok the 2nd point can be addressed by changing the default value, but the 1st point is still valid. consolidation by name means we want to initialize the model weights and synchronize them across the gpus when we call it. so making param_init_fn a True or False argument does not much sense.

comaniac: In my proposal there's no `param_init_fn` anymore, and `init_weights=[True, False, Callable]` seems not a problem to me?

Again my point is that users should call `.build` in all cases, so the `.build` API should also cover single device. Accordingly, it's not straightforward for users to understand the meaning of `param_init_fn` or `init_required` by their names. It is also easy for users to make mistake when the functionality of two arguments are tightly coupled. That's why I suggested 1) combining them to a single argument, and 2) having a better name. If `init_weights` isn't clear enough, we could come up with a better name, such as `consolidate_weights` or something like that.

szhengac: No, I am not talking about the name. What I mean is we will pass init_weights to consolidation_model func, and what is passed to consolidation_model shouldn't be a bool variable.

slapo/schedule.py
```diff
@@ -864,27 +880,25 @@ def consolidate_model(sch: Schedule, topology=None):
...
                 )
 
         # need to broadcast params from rank 0 to make sure all the TP+DP ranks take the same params
-        if not "DEBUG" in os.environ:
-            curr_stage_group = stage_groups[curr_part_idx]
-            for _, param in sch.mod.named_parameters(recurse=False):
-                dist.broadcast(param, src=curr_stage_devices[0], group=curr_stage_group)
-            # destroy process group to free memory
-            for g in stage_groups:
-                dist.destroy_process_group(g)
```
zarzen, line 887: Is the speed-up from the deletion of this line?

examples/gpt/deepspeed_hf.py
```diff
@@ -179,6 +183,8 @@ def loss_fn(outputs, labels):
     else:
         train_with_torch(model, loader, steps=num_iters)
 
+    model.save_checkpoint("checkpoint")
```
chhzh123: This line should also be removed.

slapo/autotune/tune.py
```diff
@@ -282,7 +282,7 @@ def run_training_script(args, tuneable_cfg):
     # Set all tunable parameters as environment variables.
     env = " ".join(f"{k}={v}" for k, v in tuneable_cfg.items())
 
-    cmd = f"DEBUG=1 {env} python3 {args.training_script} {train_script_args}"
+    cmd = f"{env} python3 {args.training_script} {train_script_args}"
```
chhzh123, line 285: Can we pass in the debug flag here for auto-tuning?

slapo/schedule.py
comaniac: Why not? `init_required` is also a bool variable.

tests/test_shard.py
chhzh123: Does this environment variable always exist?

slapo/autotune/tune.py
comaniac: Note that DEBUG flag is removed completely in this PR. Instead, we could probably disable consolidation using the flag in the tuning script.

chhzh123: I see

chhzh123:
LGTM. I'll approve when the interface is finalized.

slapo/schedule.py
szhengac: no, this causes nccl error

tests/test_pipeline_partition.py
```diff
@@ -51,5 +53,50 @@ def forward(self, x):
     assert ("stage2.linear.weight", 2) in tie_weights[0]
 
 
+def test_analyze_tie_ranks():
```
comaniac, line 56: I think you could just reuse the above test, as you also call `analyze_tie_weights` in this test.

slapo/schedule.py
```diff
@@ -903,29 +927,45 @@ def consolidate_model(sch: Schedule, topology=None):
...
+    sch: Schedule,
+    topology=None,
+    target=None,
+    init_weights: Optional[Union[bool, Callable]] = True,
+    **kwargs,
+):
     optimizer = None
     if sch.metadata.pipeline_cutting_paths:
+        if topology is None:
+            raise ValueError("Must provide topology for deepspeed pipeline")
```
comaniac, line 953: We should not have this check at this position because target may not be "deepspeed".

slapo/schedule.py
```diff
@@ -903,29 +927,45 @@ def consolidate_model(sch: Schedule, topology=None):
...
+
         # pipeline stages will be wrapped into PipeStageWrapper
         sch = generate_pipeline_partition(sch)
         # Analyzie tie weights before consolidation.
         tie_weight_groups = analyze_tie_weights(sch.mod)
+        tie_ranks = analyze_tie_ranks(tie_weight_groups, topology)
+        for _, _ in tie_ranks:
+            pass
```
comaniac, line 961: What's this for?

slapo/schedule.py
```diff
@@ -903,29 +927,45 @@ def consolidate_model(sch: Schedule, topology=None):
...
 
     # delay initialization
-    sch = consolidate_model(sch, topology)
+    if init_weights:
+        if isinstance(init_weights, Callable):
+            sch = consolidate_model(sch, topology, init_weights)
+        else:
+            sch = consolidate_model(sch, topology)
```
comaniac, line 968:
```suggestion
        init_weight_fn = init_weights if isinstance(init_weights, Callable) else None
        sch = consolidate_model(sch, topology, init_weight_fn)
```

comaniac (approved):
LGTM. Just nits.

tests/test_pipeline_partition.py
szhengac: not really, the above test uses 3 stages, but we can only use 2 stages when there are only 2 gpus

slapo/schedule.py
szhengac: right. i will remove deepspeed pipeline

slapo/schedule.py
szhengac: otherwise lint does not pass

tests/test_pipeline_partition.py
comaniac: Huh I just realized that this test may not pass CI because we don't have DeepSpeed installed in the docker image...can you fake a topology? This also addresses the 2 GPU issue.

slapo/schedule.py
comaniac: Interesting. What's the lint error?

szhengac: tie_ranks not used. this tie_ranks is a placeholder for you to add additional functionality

PR #6, opened 2023-01-19
[Schedule] Support sequence parallelism

comaniac:
## Description ##

- Support sequential parallelism. Specifically, now we can schedule as follows:

```python
# output would be partial
sch["attention.out_proj"].shard("weight", axis=1)
# this indicates that we sync the output in forward pass, but defer the gather
# (at output axis=1) until resid_dropout. in other words, the schedule becomes:
# linear -> reduce_scatter -> dropout -> all_gather.
sch["attention.out_proj"].sync(
    mode="forward_defer_gather",
    gather_at=(sch["attention.resid_dropout"], 1)
)
```
Note that `dist.reduce_scatter` always scatters along the first dimension, so we implicitly transpose input and output tensors when needed.

------

**UPDATE**
Per offline discussion, the programming model is changed:

```python
sch["attention.out_proj"].sync(mode="fwd_post", sync_op_or_fn="reduce_scatter", axis=1)
sch["attention.resid_dropout"].sync(mode="fwd_post", sync_op_or_fn="all_gather", axis=1)
```

Other 1: For readability and flexibility, now `.sync` requires users to always specify the op.
Other 2: The `.hook` primitive is integrated to `.sync`, which now allows a custom hook function.

------

- Accordingly, a unit test was added to verify the correctness for both forward and backward.
- Use `dist.all_gather_into_tensor` when available. This API was `dist._all_gather_base`, which will be deprecated soon. Note that `dist.all_gather_into_tensor` always concats along the first dimension, so we implicitly transpose input and output tensors when needed. Even with the transpose overheads, it seems still faster than `dist.all_gather` + `torch.cat` based on my micro benchmarks.
- Move sharding logic to another place and establish a registration mechanism for sharable modules.

- [MIsc] Refactor `task_lint.sh` so that we could run it locally to verify linting without being installed transformers everytime.
- [Misc] Add conftest.py to enforce the test order; otherwise the distributed tests may stuck if distributed devices are not running the same test at the same time.
- [Misc] Remove DeepSpeed from docker image for now, so that we could make it public for CI.

cc @szhengac @chhzh123 

## Checklist ##

- [x] Code is well-documented


examples/opt/schedule.py
```diff
@@ -197,7 +197,10 @@ def pattern(x: torch.Tensor) -> torch.Tensor:
             sub_sch["module.FusedQKV_0.fused_linear"].shard("bias", axis=0)
             sub_sch["module.FusedQKV_0.fused_linear"].sync(mode="backward")
             sub_sch["module.out_proj"].shard("weight", axis=1)
-            sub_sch["module.out_proj"].sync(mode="forward")
+            sub_sch["module.out_proj"].sync(
+                mode="forward_defer_gather",
+                gather_at=(sub_sch["module.resid_dropout"], 1),
```
chhzh123, line 202: Can we pass in two kwargs instead of a tuple? The number without a name looks weird.

slapo/schedule.py
```diff
@@ -311,24 +386,53 @@ def sync(self, mode="backward"):
...
+                        raise ValueError(
+                            "gather_at is missing in kwargs for mode "
+                            "forward_defer_gather"
+                        )
+
+                    # Case 4
+                    gather_sch, gather_axis = kwargs["gather_at"]
+                    sync_fn = partial(
+                        reduce_scatter_forward_output, dim=gather_axis, group=self.group
+                    )
```
chhzh123, line 423: Do we need to verify whether the gather_axis is valid?

comaniac: It is verified in `reduce_scatter_forward_output`. The only requirement is the dimension being scattered has to be divisible by TP size.

examples/opt/schedule.py
comaniac: This is refactored per offline discussion with @szhengac.

slapo/schedule.py
```diff
@@ -311,14 +266,47 @@ def sync(self, mode="backward"):
         Case 3: (shard x, shard_in w) -> partial sum -> allreduce
                 -> (replica x, shard_out w).
             In this case, backward does not need allreduce, so mode should be 'forward'.
+        Case 4: (shard x, shard_in w) -> partial sum -> reduce-scatter
+                -> ... -> allgather -> full output.
+            This case breaks the allreduce in case 3 to reduce-scatter and allgather,
+            which is called "sequential parallelism". In this case, we also need
```
szhengac, line 272: sequence

slapo/schedule.py
```diff
@@ -311,14 +266,47 @@ def sync(self, mode="backward"):
...
+            to specify the allgather point in kwargs. For example,
+            sch["out_proj"].sync(mode="forward_defer_gather", gather_at=sch)
```
szhengac, line 274: the api has changed

slapo/schedule.py
```diff
@@ -341,6 +329,8 @@ def hook_func(_module, _input, output):
 
         if mode in {"backward", "both"}:
             # Case 1, 2
+            if sync_op is not None:
+                raise ValueError("sync_op is not supported in backward yet")
```
szhengac, line 333: i am a bit confused why sync_op is not supported for backward. Don't we need corresponding op for reduce scatter and allgather?

slapo/sharding/utils.py
```diff
@@ -0,0 +1,152 @@
...
+        return all_gather_along_dim(inp, dim, rank, world_size, group)
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        dim = ctx.dim
+        group = ctx.group
+        world_size = dist.get_world_size(group)
+        rank = dist.get_rank(group)
+        sharded_size = grad_output.shape[dim] // world_size
+        ret = grad_output.split(sharded_size, dim=dim)[rank]
```
szhengac, line 72: we would need contiguous when dim is the primary dimension.

slapo/schedule.py
comaniac: See the implementation of all_gather_forward_output.

szhengac: i see. so they have been coupled. then, why don't we use the same rule for allreduce since backward behavior is always coupled with forward.

slapo/schedule.py
comaniac: Good catch.

slapo/sharding/utils.py
comaniac: Fixed.

slapo/schedule.py
comaniac: Because it's not always coupled.

Case 1: Partitioned output with all-gather.
```
Forward: all-gather
Backward: split -> allreduce.
```
In this case, all-gather and split are coupled, so we use `all_gather_forward_output`, but backward needs a standalone allreduce. We use `sch.sync(mode="both")`.

Case 2: Partitioned output without all-gather.
```
Forward: N/A
Backward: allreduce.
```
In this case, backend needs a standalone allreduce, so we use `sch.sync(mode="backward")`.

Case 3: Partial sum with allreduce.
```
Forward: allreduce
Backward: Identical
```
In this case, backward doesn't need anything, so we just `sch.sync(mode="forward")`.

szhengac (approved):

conftest.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def init_dist(request):
+    """Initialize the distributed group once in the entire test session."""
+    torch.manual_seed(9999)
+    try:
+        dist.init_process_group(backend="nccl")
+    except Exception as err:
+        print(f"Skip === {str(err)}")
```
comaniac, line 25:
```suggestion
```

examples/gpt/schedule.py
```diff
@@ -168,9 +168,16 @@ def pattern(x: torch.Tensor) -> torch.Tensor:
...
             sub_sch["module.FusedQKV_0.fused_linear"].shard("bias", axis=0)
-            sub_sch["module.FusedQKV_0.fused_linear"].sync(mode="backward")
+            sub_sch["module.FusedQKV_0.fused_linear"].sync(
+                mode="bwd_post", sync_op_or_fn="all_reduce"
+            )
             sub_sch["module.out_proj"].shard("weight", axis=1)
-            sub_sch["module.out_proj"].sync(mode="forward")
+            sub_sch["module.out_proj"].sync(
+                mode="fwd_post", sync_op_or_fn="reduce_scatter", axis=0
+            )
```
szhengac, line 177: I thought it should be axis=1? axis=1 is the sequence dimension?

examples/gpt/schedule.py
```diff
@@ -168,9 +168,16 @@ def pattern(x: torch.Tensor) -> torch.Tensor:
...
+            sub_sch["module.resid_dropout"].sync(
+                mode="fwd_post", sync_op_or_fn="all_gather", axis=0
```
szhengac, line 179: similar

slapo/schedule.py
```diff
@@ -276,109 +249,165 @@ def set_output_type(output_type, gather_axis=None):
...
 
         Case 1: (replica x, shard_out w) -> partition output -> allgather
                 -> full output -> (replica x, shard_out w).
             In this case, since forward uses all-gather to get a full output,
             backward must have a split to match the shape, and
-            allreduce is also required for x.grad, so mode should be 'both'.
+            allreduce is also required for x.grad, so we use:
+            ```python
+            sch["out_prj"].shard("weight", axis=0)
+            sch["out_prj"].sync(mode="fwd_post", sync_op_or_fn="all_gather", axis=1)
```
szhengac, line 272: i found it might a bit confusing since shard(axis=0) is for weight output dimension but sync(axis=1) for activation output dimension. maybe add a bit more text to explain?

examples/gpt/schedule.py
comaniac: Oops, good catch.

examples/albert/schedule.py
```diff
@@ -232,24 +234,24 @@ def shard_word_embedding(sch, vocab_size, word_embed_name="embeddings.word_embed
     vocab_start_index = sch.rank * vocab_size // sch.world_size
     vocab_end_index = (sch.rank + 1) * vocab_size // sch.world_size
 
-    def fw_pre_hook(_input):
+    def fwd_pre_hook(_module, _input):
         # Mask the input
         input_mask = (_input[0] < vocab_start_index) | (_input[0] >= vocab_end_index)
```
szhengac, line 239: btw, _input here has shape (B, T), so we should just check `_input < vocab_start_index) | (_input >= vocab_end_index)`, otherwise `_input[0] is only for the first example`?

comaniac: IIUC this input is always a list, so input[0] should be the entire tensor.

szhengac: wait, if this is a list, why we only return the masked_input? the module should expect the same input structure?

comaniac: That's because this case only has one input tensor. In this case pytorch will wrap a single returned tensor to a tuple, according to https://pytorch.org/docs/stable/generated/torch.nn.modules.module.register_module_forward_pre_hook.html

szhengac: ok makes sense

comaniac:
Thanks @szhengac

PR #8, opened 2023-01-20
[Pipeline] Tie weight analysis

comaniac:
## Description ##
This PR analyzes whether parameters are tie together by looking at their `nn.Paramter` object IDs. Accordingly, this analysis has to be done before the consolidation (if needed); otherwise all `nn.Parameter` objects will be replaced so that their object IDs won't be the same anymore.

Note that the above analysis works only when parameters are pointing to the same Python object. In other words, for the distributed training that requires weight consolidation, users have to explicitly call `tie_weights` one more time when initializing the model with empty weights. For example:

```python
with slapo.init_empty_weights():
    model = get_model()
    model.tie_weights()

sch = slapo.create_schedule(model, ...)
# The .build does the following in order:
# 1. Partition pipeline stages.
# 2. Tie weight analysis.
# 3. Consolidation (this wipes out the previous tie weights in the model)
# 4. Liveness analysis.
# 5. Pipeline module construction (this makes use the tie weight and liveness analysis results)
sch.build(...)
```

In addition, this PR also refactors the pipeline logic in `schedu.build`. The ultimate goal is to ensure that `schedule.build` doesn't have any framework specific logic (e.g., `if target == "deepspeed":`).

cc @szhengac @zarzen 

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/slapo/initialization.py
+++ b/slapo/initialization.py
@@ -59,7 +59,7 @@ def init_on_device(device: torch.device, include_buffers: bool = False):
 
     def register_empty_parameter(module, name, param):
         old_register_parameter(module, name, param)
-        if param is not None:
+        if param is not None and param.device != device:
             param_cls = type(module._parameters[name])
             kwargs = module._parameters[name].__dict__
             module._parameters[name] = param_cls(
--- a/slapo/model_dialect/deepspeed/pipeline.py
+++ b/slapo/model_dialect/deepspeed/pipeline.py
@@ -295,6 +295,7 @@ class DeepSpeedPipeStageWrapper(nn.Module):
         ret = tuple(ret)
         return ret
 
+
 @register_model_dialect("deepspeed", "pipeline_engine")
 def deepspeed_pipe_engine(
     stage_modules,
--- a/slapo/model_dialect/deepspeed/pipeline.py
+++ b/slapo/model_dialect/deepspeed/pipeline.py
@@ -1,5 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
 
 from enum import Enum
 import torch
@@ -168,7 +169,7 @@ def unflatten(args, metadata):
     return tupleize(unordered_args)
 
 
-@register_model_dialect("deepspeed", "pipeline")
+@register_model_dialect("deepspeed", "pipeline_stage")
 class DeepSpeedPipeStageWrapper(nn.Module):
     def __init__(
         self,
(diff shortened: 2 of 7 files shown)


slapo/initialization.py
```diff
@@ -59,7 +59,7 @@ def init_on_device(device: torch.device, include_buffers: bool = False):
 
     def register_empty_parameter(module, name, param):
         old_register_parameter(module, name, param)
-        if param is not None:
+        if param is not None and param.device != device:
```
szhengac, line 62: what is this used for? I think register_empty_parameter is recalled after we exit the init context?

slapo/pipeline.py
```diff
@@ -274,6 +275,78 @@ def get_itemized_name(node, suffix=""):
...
+                            target_param_full_name
+                        ] = target_param_full_name
+                        tie_group = set([(target_param_full_name, target_stage)])
+                        tie_groups[target_param_full_name] = tie_group
+
+                    # Add the current parameter to the tie group.
+                    param_name_2_group_key[
+                        curr_param_full_name
+                    ] = target_param_full_name
+                    tie_group.add((curr_param_full_name, stage_id))
```
szhengac, line 333: Set is missing

slapo/initialization.py
comaniac: Without this, wte.weight = lm_head.weight won't let them have the same parameter object.

slapo/pipeline.py
comaniac: What set?

szhengac: tie_group value is a Set based on your docstring and the other part of the code.

slapo/initialization.py
szhengac: ok so this is triggered during init?

comaniac: No it's triggered by `__setattr__` -> `register_parameter`

slapo/pipeline.py
comaniac: So tie_group.add adds a new value to the set. I didn't see the problem?

szhengac: Oh nvm, I look at the wrong variable.

szhengac:
LGTM.

szhengac (approved):

szhengac:
Let me know if it is ready for merge

comaniac:
Thanks @szhengac

slapo/model_dialect/deepspeed/pipeline.py
```diff
@@ -293,3 +294,25 @@ def forward(self, *args, **kwargs):
...
+def deepspeed_pipe_engine(
+    stage_modules,
+    topology,
+    param_dtype,
+    **kwargs,
+):
+    from deepspeed import pipe
+
+    model = pipe.PipelineModule(
+        stage_modules,
```
zarzen, line 309: Is `stage_modules` a list of layers? Or is it partitioned by the Slapo schedule?

PR #9, opened 2023-01-20
[Bugfix] Reproduce experimental results in docker image

chhzh123:
## Description ##
This PR tries to reproduce the experimental results presented in the paper using the provided docker image. Most of the errors are dependency issues.

## Checklist ##

Following items list the failed test cases and the root causes:
- [x] slapo-megatron, albert, gpu=1,2,4,8 (`networkx` is required by functorch, `epoi` must use the lastest commit to support the injection policy)
- [x] slapo-deepspeed, albert, gpu=2,4,8 (same issue)
- [x] slapo-megatron, gpt, gpu=2,4,8 (same issue)
- [x] deepspeed, gpt, gpu=2,4,8 (missing `datasets` library)
- [x] slapo-deepspeed, gpt, gpu=2,4,8 (same issue)
- [x] slapo-megatron, wideresnet, gpu=2,4,8 (TorchScript is used in multi-device)
- [x] slapo-deepspeed, wideresnet, gpu=2,4,8 (same issue)

### DIFF
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -15,7 +15,7 @@ def get_ds_config(
     batch_size, micro_batch_size_per_gpu, fp16=True, zero3=False, desc="", bf16=False
 ):
     # https://github.com/microsoft/DeepSpeed/blob/ff42743/tests/unit/model_parallelism/test_configurable_parallel_pp.py#L20
-    assert fp16 ^ bf16, f"Either fp16={fp16} or bf16={bf16} needs to be turned on"
+    logger.info(f"fp16={fp16}, bf16={bf16}")
     config_dict = {
         "help": desc,
         "steps_per_print": 10,
--- a/examples/wideresnet/model.py
+++ b/examples/wideresnet/model.py
@@ -53,8 +53,6 @@ def schedule_model(
     sch = slapo.create_schedule(model, group=group)
     logger.info(f"Scheduling Wide-ResNet with TP={sch.world_size}", ranks=0)
 
-    fuse_conv_bn(sch[prefix], config)
-
     if sch.world_size > 1:
         # Shard layers.
         shard_layers(sch[prefix], config)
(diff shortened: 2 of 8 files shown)


chhzh123:
Has run all the end-to-end single-device and single-node experiments in the docker and verified the results can be reproduced.

examples/wideresnet/model.py
```diff
@@ -53,7 +54,8 @@ def schedule_model(
     sch = slapo.create_schedule(model, group=group)
     logger.info(f"Scheduling Wide-ResNet with TP={sch.world_size}", ranks=0)
 
-    fuse_conv_bn(sch[prefix], config)
+    if dist.get_world_size() == 1:  # not only counting the TP devices
```
comaniac, line 57: Use `sch.world_size` would be our best practice to make use this is for a TP group.
```suggestion
    if sch.world_size == 1:  # not only counting the TP devices
```

comaniac (changes requested):
I'll update the docker image accordingly.

comaniac:
Docker image updated.

examples/utils.py
```diff
@@ -15,7 +15,7 @@ def get_ds_config(
     batch_size, micro_batch_size_per_gpu, fp16=True, zero3=False, desc="", bf16=False
 ):
     # https://github.com/microsoft/DeepSpeed/blob/ff42743/tests/unit/model_parallelism/test_configurable_parallel_pp.py#L20
-    assert fp16 ^ bf16, f"Either fp16={fp16} or bf16={bf16} needs to be turned on"
```
comaniac, line 18: This assertion is removed. Is this intentional?

slapo/schedule.py
```diff
@@ -873,6 +873,11 @@ def _init_module(sch: Schedule):
             )
 
     def _consolidate_and_broadcast(sch: Schedule):
+        if isinstance(sch.mod, torch.jit.ScriptModule):
+            # Scripted module requires the parameters have been initialized,
+            # and the original methods are not preserved, so no need to consolidate
```
comaniac, line 878:
```suggestion
            # Scripted module requires the parameters to be initialized in advance,
            # so no need to consolidate
```

comaniac (approved):
Otherwise LGTM

examples/utils.py
chhzh123: Yes, WideResNet uses fp32 and cannot pass this assertion

PR #16, opened 2023-01-24
[Test] Add tracer unit tests

chhzh123:
## Description ##
This PR updates the unit tests for tracer. Also, it adds a syntactic sugar for users to easily retrieve submodules in `nn.Sequential`. For example, previously, users can only use `sch["layer1"]["0"]["downsample"]` to access submodules in WideResNet. This PR enables the `sch["layer1.0.downsample"]` access.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/slapo/schedule.py
+++ b/slapo/schedule.py
@@ -193,21 +193,17 @@ class Schedule:
 
         curr_sch = self
         for token in self.tokenize_module_path(full_path):
-            try:  # nn.Sequential
-                new_sch = curr_sch
-                for subtoken in token.split("."):
-                    if subtoken not in new_sch.child:
-                        raise KeyError(
-                            f"Try to find nn.Sequential modules, but the schedule of '{full_path}' ({subtoken}) is not a child of {new_sch.name}"
-                        )
-                    new_sch = new_sch.child[subtoken]
-                curr_sch = new_sch
-            except Exception as exc:  # nn.Module/nn.ModuleList
-                if token not in curr_sch.child:
-                    raise KeyError(
-                        f"The schedule of '{full_path}' ({token}) is not a child of {curr_sch.name}"
-                    ) from exc
-                curr_sch = curr_sch.child[token]
+            sub_tokens = token.split(".")
+            if len(sub_tokens) == 2 and sub_tokens[0] in curr_sch.child:
+                # If this token is in the format of "layer.0" and "layer" is a child of curr_sch,
+                # then "layer" is nn.Sequential. In this case, we have to first get the nn.Sequential module first.
+                curr_sch = curr_sch.child[sub_tokens[0]]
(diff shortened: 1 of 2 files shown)


slapo/schedule.py
```diff
@@ -173,11 +173,21 @@ def __getitem__(self, full_path):
 
         curr_sch = self
         for token in self.tokenize_module_path(full_path):
-            if token not in curr_sch.child:
-                raise KeyError(
-                    f"The schedule of '{full_path}' is not a child of {curr_sch.name}"
-                )
-            curr_sch = curr_sch.child[token]
+            try:  # nn.Sequential
+                new_sch = curr_sch
```
comaniac, line 177: `new_sch` is a bit misleading. `sub_sch` or `child_sch` might be better.

slapo/schedule.py
```diff
@@ -173,11 +173,21 @@ def __getitem__(self, full_path):
...
+                            f"Try to find nn.Sequential modules, but the schedule of '{full_path}' ({subtoken}) is not a child of {new_sch.name}"
+                        )
+                    new_sch = new_sch.child[subtoken]
+                curr_sch = new_sch
+            except Exception as exc:  # nn.Module/nn.ModuleList
+                if token not in curr_sch.child:
+                    raise KeyError(
+                        f"The schedule of '{full_path}' ({token}) is not a child of {curr_sch.name}"
+                    ) from exc
+                curr_sch = curr_sch.child[token]
```
comaniac, line 190: This implementation is a bit hard to read and could be simplified. For example, I don't think you need a for-loop for `for subtoken in token.split(".")`, because the token from `tokenize_module_path` include at most one `.` (i.e., `module.index`). So overall I think we could get rid of the try-catch block and have more comments:

```python
sub_tokens = tokens.split(".")
if len(sub_tokens) == 2 and sub_tokens[0] in curr_sch.child:
    # If this token is in the format of "layer.0" and "layer" is a child of curr_sch,
    # then "layer" is nn.Sequential. In this case, we have to first get the nn.Sequential module first.
    curr_sch = curr_sch.child[sub_tokens[0]]
    token = sub_token[1]
if token not in curr_sch.child:
    raise KeyError(
        f"The schedule of '{full_path}' ({token}) is not a child of {curr_sch.name}"
    )
curr_sch = curr_sch.child[token]
```

comaniac (commented):
Otherwise LGTM. You could add a new commit to trigger the CI. It should work now.

slapo/schedule.py
chhzh123: Changed accordingly. Thanks for the suggestion.

comaniac (approved):
LGTM

comaniac:
Thanks @chhzh123

PR #18, opened 2023-01-25
[Release] Setup wheel and release scripts

comaniac:
## Description ##
- Setup scripts for building wheels and uploading to PYPI.
- Change the versioning approach. Now we follow Alpa and Ray to use PEP-440 convention.
- Add a Github Action workflow to automatically build and upload the latest wheel to PYPI when we make a new release in the repo. There's no way to test this workflow right now, so we need to keep an eye on the next release.
- When packing, I found that some folders don't have `__init__.py` so that will be ignored by `sdist` and `pylint`.
- Accordingly, this PR also fixes some more lint errors that we missed previously.

The first release `v0.0.1` cut at the current main branch has been uploaded to PYPI. We will keep using `v0.0` series for now. When everything is ready, including unit tests, documents, etc, we will release `v0.1`, which should be the first official public release of Slapo.

## Checklist ##

- [x] Code is well-documented

cc @chhzh123


chhzh123:
LGTM. Thanks @comaniac for setting up the wheel.

PR #23, opened 2023-01-27
[Schedule] Fix linear bias after row sharding weight (#10)

### LINKED ISSUE #10: [Bug] Multiple reductions of linear bias after row sharding the weight
Consider the following case, we have partial inputs and shard the weight $A$ of the linear layer by row.

$$
XA+b
=\begin{bmatrix}X_1 & X_2\end{bmatrix}
\begin{bmatrix}A_1 \\\\ A_2\end{bmatrix}
+b
= X_1A_1+X_2A_2+b
$$

In the current implementation (as shown below), we directly conduct `all_reduce` for the output of the linear layer. However, since the bias $b$ is implicitly included in the linear layer, this leads to redundant additions for the bias and generates incorrect results.

```python
class Model(nn.Module):
  def __init__(self):
    self.linear = nn.linear(...)
  def forward(self, x):
    x = self.linear(x)
    dist.all_reduce(x, op=dist.ReduceOp.SUM)
    return x
```

As a result, we need to preprocess the bias when the corresponding linear layer is row sharded. There are two possible solutions:
1. Only one device holds the bias data and other devices set bias to None, so the bias will only be added once.
2. Average the bias across different devices. In that case, the summation of the new bias will be the same as the original one.

chhzh123:
## Description ##
This PR fixes #10. It leverages Megatron-LM's method to delay the bias add after `all_reduce`, so that the bias will only be added once, and the gradients are proved to be the same as the original ones.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/slapo/op/linear.py
+++ b/slapo/op/linear.py
@@ -19,22 +19,24 @@ class LinearWithSeparateBias(nn.Linear):
         self,
         in_features: int,
         out_features: int,
-        world_size: int,
-        group: dist.ProcessGroup,
+        sync_fn,
         device=None,
         dtype=None,
     ) -> None:
         factory_kwargs = {"device": device, "dtype": dtype}
-        super(LinearWithSeparateBias, self).__init__(
+        super().__init__(
             in_features, out_features, bias=False, device=device, dtype=dtype
         )
-        self.world_size = world_size
-        self.group = group
+        self.sync_fn = sync_fn
         self.bias = Parameter(torch.empty(out_features, **factory_kwargs))
         self.reset_parameters()
 
-    def forward(self, x: Tensor) -> Tensor:
+    def forward(self, x: Tensor) -> Tensor:  # pylint: disable=arguments-renamed
         x = F.linear(x, self.weight, None)
-        dist.all_reduce(x, op=dist.ReduceOp.SUM, group=self.group)
+        # pylint: disable=comparison-with-callable
+        if self.sync_fn.func == dist.all_reduce:
+            self.sync_fn(x)
+        else:
+            x = self.sync_fn(x)
         x = x + self.bias
         return x
--- a/slapo/op/linear.py
+++ b/slapo/op/linear.py
@@ -4,6 +4,7 @@
 import torch
 from torch import nn
 import torch.nn.functional as F
+import torch.distributed as dist
 from torch import Tensor
 from torch.nn.parameter import Parameter
 
@@ -19,6 +20,7 @@ class LinearWithSeparateBias(nn.Linear):
         in_features: int,
         out_features: int,
         world_size: int,
+        group: dist.ProcessGroup,
         device=None,
         dtype=None,
     ) -> None:
@@ -27,11 +29,12 @@ class LinearWithSeparateBias(nn.Linear):
             in_features, out_features, bias=False, device=device, dtype=dtype
         )
         self.world_size = world_size
+        self.group = group
         self.bias = Parameter(torch.empty(out_features, **factory_kwargs))
         self.reset_parameters()
 
     def forward(self, x: Tensor) -> Tensor:
         x = F.linear(x, self.weight, None)
-        # Divide bias by world_size to avoid redundant summation in TP
-        x = x + self.bias / self.world_size
+        dist.all_reduce(x, op=dist.ReduceOp.SUM, group=self.group)
+        x = x + self.bias
         return x
--- a/slapo/op/linear.py
+++ b/slapo/op/linear.py
@@ -1,0 +1,37 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from torch import nn
+import torch.nn.functional as F
+from torch import Tensor
+from torch.nn.parameter import Parameter
+
+
+class LinearWithSeparateBias(nn.Linear):
+    """Implementation modified from `nn.Linear`
+    This class is also inherited from `nn.Linear` to make sure it will be
+    treated as the same class when using `isinstance(...)` in `init_weights`
+    """
+
+    def __init__(
+        self,
+        in_features: int,
+        out_features: int,
+        world_size: int,
+        device=None,
+        dtype=None,
+    ) -> None:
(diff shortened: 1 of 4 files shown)


chhzh123:
But the sequence parallelism case have not passed yet. I'm not sure why `linear2.bias.grad` is 1/TP of the original one, though I only add the bias once.

slapo/op/linear.py
```diff
@@ -0,0 +1,42 @@
...
+        factory_kwargs = {"device": device, "dtype": dtype}
+        super().__init__(
+            in_features, out_features, bias=False, device=device, dtype=dtype
+        )
+        self.sync_fn = sync_fn
+        self.bias = Parameter(torch.empty(out_features, **factory_kwargs))
+        self.reset_parameters()
+
+    def forward(self, x: Tensor) -> Tensor:  # pylint: disable=arguments-renamed
+        x = F.linear(x, self.weight, None)
```
comaniac, line 35: Can you just call the forward function of the base class (i.e., linear)? I suppose it should have the same functionality, but might be more clean.

slapo/op/linear.py
```diff
@@ -0,0 +1,42 @@
...
+        # pylint: disable=comparison-with-callable
+        if self.sync_fn.func == dist.all_reduce:
+            self.sync_fn(x)
+        else:
+            x = self.sync_fn(x)
```
comaniac, line 40: Add a comment saying what this sync function could be. It would be good to also add a complete docstring including arguments for this class.

slapo/schedule.py
```diff
@@ -428,6 +430,26 @@ def validate_sync_op(mode, sync_op_or_fn, axis=None):
...
+                        ):
+                            new_mod = LinearWithSeparateBias(
+                                self.mod.in_features,
+                                self.mod.out_features,
+                                sync_fn,
+                                self.mod.weight.device,
+                                self.mod.weight.dtype,
+                            )
+                            self.replace(new_mod)
+                        return
```
comaniac, line 452: Some concerns:
1. This breaks tie weights.
2. The execution flow of .shard becomes a bit difficult to flow, because here is an early return. I'm not sure if there's a better way but I'll think about it.

szhengac:
> But the sequence parallelism case has not passed yet. I'm not sure why `linear2.bias.grad` is 1/TP of the original one, though I only add the bias once.

As I mentioned offline, this is because each bias only pass through 1/TP tokens due to sequence dimension partitioning. We need to discuss it more before we can merge this pr.

slapo/op/linear.py
chhzh123: I removed the previous bias creation and `reset_parameters`, but retain the `F.linear` function. Should have the same functionality.

slapo/op/linear.py
chhzh123: Fixed

slapo/schedule.py
chhzh123: Added `tie_weights` support

slapo/schedule.py
```diff
@@ -407,6 +409,44 @@ def validate_sync_op(mode, sync_op_or_fn, axis=None):
...
+                            "but they have different sharded shapes: "
+                            f"{new_tensor.shape} vs "
+                            f"{sch.metadata.tie_weights[param].shape}"
+                        )
+                    new_param = sch.metadata.tie_weights[param]
+                else:
+                    new_param = new_mod.weight
+                    self.metadata.tie_weights[param] = new_param
+            else:
+                new_param = new_mod.weight
```
comaniac, line 447: Could we have a utility function for this logic so that we could just call it when needed?

chhzh123: Added

chhzh123:
> But the sequence parallelism case has not passed yet. I'm not sure why `linear2.bias.grad` is 1/TP of the original one, though I only add the bias once.

Extra [parameter hook](https://pytorch.org/docs/stable/generated/torch.Tensor.register_hook.html) is added to make sure the bias grad is correct.

slapo/schedule.py
```diff
@@ -219,6 +224,24 @@ def __contains__(self, full_path):
                 return False
         return True
 
+    def get_params_if_tied(self, old_param, new_param):
```
comaniac, line 227: The function name seems not clear to me. Maybe "get_or_set_tied_param" or something like that.

slapo/schedule.py
```diff
@@ -1093,6 +1142,18 @@ def init_target_engine(sch, target, **kwargs):
     )
 
 
+def reduce_model_grads(sch: Schedule):
```
comaniac, line 1145: I like this idea, but not sure if there's any issue with this implementation. Maybe we could revisit it later.

comaniac (approved):
LGTM. Just nits. I'll leave this PR to @szhengac for merging.

slapo/schedule.py
```diff
@@ -1154,6 +1160,18 @@ def init_target_engine(sch, target, **kwargs):
...
+    def hook(grad):
+        dist.all_reduce(grad.contiguous(), dist.ReduceOp.SUM, group=sch.group)
```
szhengac, line 1147: This is not an efficient way to allreduce bias, since there are hundreds of bias in a large model. and we will end up with large communication latency, greatly slowing down the training. The best way for bias will be merging them in the end to perform a single allreduce call. for example, take a look at where layernorm gradient is reduced in megatron https://github.com/NVIDIA/Megatron-LM/blob/master/megatron/optimizer/optimizer.py#L285

chhzh123: Right, I've seen the Megatron implementation, but the problem is that it is implemented in optimizer, so it can all_reduce the parameters in one place. For our case, we only return a model for users, and the computation and communication should be included in the model, so that's why I use param hooks instead.

szhengac: you can invoke gradient sync after the whole stage backward. nonetheless, we cannot use the current way of synchronizing gradients otherwise we will end up with extremely slow training example.

slapo/schedule.py
chhzh123: Fixed

chhzh123:
Moved gradient reduction to the end of each backward pass per @szhengac 's suggestion.

comaniac:
Closed as #74 already covers this feature.

PR #25, opened 2023-01-29
[Schedule] Add .fuse() primitive

chhzh123:
## Description ##
This PR adds a new primitive called `.fuse(subgraph, compiler)` for operator fusion. Currently we only support pattern-based [vertical operator fusion](https://sisyphus.gitbook.io/project/deep-learning-basics/computation-graph-optimization/tensorrt-computation-graph-optimization) using TorchScript as the backend compiler. A simple example is shown below.

```python
class Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.conv = nn.Conv2d(3, 32, 3)

    def forward(self, x):
        x = self.conv(x)
        x = F.relu(x)
        x = x + 1
        return x

mod = Model()
sch = slapo.create_schedule(mod)

# Define a function pattern whose ops will be fused
def pattern(x: torch.Tensor):
    x = F.relu(x)
    x = x + 1
    return x

# Find the subgraph in the original module that matches the specified pattern
subgraph = sch.find(pattern)
# Fuse the subgraph
sch.fuse(subgraph, compiler="TorchScript", name="FusedReLU")
```

The fusion performance needs to be tested. As I first create an identical torch.fx subgraph and pass it into TorchScript's scripting mode for optimization, it may cause performance issue if some of the operators are not recognized by TorchScript. Tracing mode can achieve the best performance but it is hard to leverage since we cannot always obtain example inputs for each subgraph. Scripting a user-defined function is also not a good approach since the backward pass is not captured by TorchScript. Therefore, only scripting an entire module is a good fit for our case, and we need to test the compatibility of torch.fx and TorchScript.

## Checklist ##

- [x] Support using op in `torch.nn.functional` to match modules in `torch.nn`. For example, `F.relu` and `nn.ReLU` should be treated as the same in pattern matching, since users cannot specify a module in a function pattern. (#35)
- [x] Support subgraph matching with multiple input arguments. (#35)
- [x] Add `.decompose()` primitive and support decoupling bias from `nn.Linear`
- [x] Support generate a flattened traced graph (#29)
- [x] Fuse the decomposed bias with consequential operators, e.g. bias+GeLU.

## Future plan (Updated) ##
The following features will be added in separate PRs.
* Subgraph matching with multiple input arguments (#35)
* Extensible compiler backend (e.g., TVM)
* Rule-based subgraph matching (#35), probably needs more use cases and testings

### DIFF
--- a/examples/t5/schedule.py
+++ b/examples/t5/schedule.py
@@ -141,7 +141,7 @@ def replace_and_shard_attention(
                 def forward(self, hidden_states):
                     states = self.query(hidden_states)
                     states = self.reshape_for_scores(states)
-                    return [states]
+                    return states
 
             def pattern_kv(x: torch.Tensor) -> torch.Tensor:
                 x = call_module("key|value", x)
@@ -162,7 +162,7 @@ def replace_and_shard_attention(
                 return x
 
             subgraphs = sub_sch.find(pattern_q)
-            assert len(subgraphs) != 0
+            assert len(subgraphs) == 1
             with init_empty_weights(enable=delay_init):
                 new_q = ShardableQ(num_heads, hidden_size, d_kv)
             sub_sch.replace(new_q, subgraphs)
--- a/slapo/pattern.py
+++ b/slapo/pattern.py
@@ -9,5 +9,14 @@ class Pattern(nn.Module):
         raise NotImplementedError
 
 
+class CallModule(nn.Module):
+    def __init__(self, name):
+        super().__init__()
+        self.name = name
+
+    def forward(self, *args):
+        raise NotImplementedError
+
+
 def call_module(mod_name, *args):
     raise NotImplementedError
--- a/slapo/pattern.py
+++ b/slapo/pattern.py
@@ -9,7 +9,7 @@ class Pattern(nn.Module):
         raise NotImplementedError
 
 
-class CallModule(nn.Module):
+class ModulePattern(nn.Module):
     def __init__(self, name):
         super().__init__()
         self.name = name
--- a/examples/albert/schedule.py
+++ b/examples/albert/schedule.py
@@ -10,6 +10,7 @@ import torch.nn as nn
 from slapo import init_empty_weights
 from typing import Optional
 from slapo.pattern import call_module
+from slapo.op.linear import FusedQKV
 
 
 def trace_attention(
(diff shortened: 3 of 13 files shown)


comaniac:
Seems like we could ask for sample inputs in kwargs if required by the compiler backend. For example: `compiler="torchscript", example_inputs=...`

chhzh123:
It seems this PR requires several changes to the subgraph matching mechanism. I will open a new PR to make step 1 & 2 work first.

epoi
```diff
@@ -0,0 +1 @@
+Subproject commit b2e2e986bc6173b60ac87dab646d9ce66e954742
```
comaniac, line 1: Why have this?

examples/t5/schedule.py
```diff
@@ -141,7 +141,7 @@ def reshape_for_scores(self, x):
                 def forward(self, hidden_states):
                     states = self.query(hidden_states)
                     states = self.reshape_for_scores(states)
-                    return [states]
+                    return states
```
comaniac, line 144: Is this a bug I did before?

slapo/pattern.py
```diff
@@ -9,5 +9,14 @@ def forward(self, *args):
         raise NotImplementedError
 
 
+class CallModule(nn.Module):
```
comaniac, line 12: The names of CallModule and call_module are confusing.

comaniac (approved):
Overall LGTM. Later we should apply .fuse to the examples, which currently replace an entire MLP for bias_new_gelu.

examples/t5/schedule.py
chhzh123: It is actually a bug of `replace_module`. See these several lines of code:
https://github.com/awslabs/slapo/blob/3393bdfbf8c60370736906381c6111cfc16369a0/slapo/schedule.py#L770-L772
Currently `subgraphs` is a list of list of tuple, which means the if-condition is always true. After fixing the condition, `pattern_q` should go through the vertical fusion branch, and thus the return of the replaced module is just a tensor instead of a list.

slapo/pattern.py
chhzh123: They are just used in different places. `CallModule` is used in a pattern class, while `call_module` can be used in a pattern function. Do you have any suggestions about the naming?

chhzh123:
> Later we should apply .fuse to the examples, which currently replace an entire MLP for bias_new_gelu.

Sure, I'll add it in the next PR.

slapo/pattern.py
comaniac: Hmm base on your use cases, such as

```python
    class LinearReLUPattern(slapo.Pattern):
        def __init__(self):
            super().__init__()
            self.fc = CallModule(r"fc?")
            self.relu = nn.ReLU()
```

Here you're actually constructing a module instead of calling it, so I guess you could use the name like `ModulePattern` or something like that.

chhzh123: It may also lead to confusion with the `Pattern` class.

comaniac: Well, to me it is a kind of pattern, especially you have a regex in its argument.

chhzh123: Yes, I agree it is a pattern, and `ModulePattern` is a proper name, but since we have the `Pattern` base class which is also a module, I'm afraid it will cause confusion between these two.

comaniac:
Thanks @chhzh123

Issue #24, opened 2023-01-28
[Bug] Original hooks are discarded in pipeline module


chhzh123:
Since we replace the top-level module with the pipeline-wrapped DeepSpeed module, the hooks attached to the original module are discarded (e.g., broadcast inputs) leading to unintended behaviors. This issue does not only happen in pipeline module, but also appears when the original module is replaced, so we need to re-hook the functions to guarantee the correctness.

PR #26, opened 2023-01-30
[Bugfix] Fix GPT script

szhengac:
## Description ##
This pr fixed the following issues:
1. Data broadcast is not hooked to submod_0 after pipeline partition and pipeline module creation. Also, rank 0 is not the correct src rank for broadcast, where it take as input a global rank. This pr provides a temporary fix, while a long term solution is tracked in https://github.com/awslabs/slapo/issues/24.
2. Data samples in the first and last stages are not consistent causing output-label mismatch
3. Embedding sharding post-hook cannot directly use an all-reduce collective since it does not handle the backward. I added an AutoGrad function to enable identity mapping for the backward.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/examples/albert/deepspeed_hf.py
+++ b/examples/albert/deepspeed_hf.py
@@ -112,7 +112,7 @@ def train(args):
             lm_loss = lm_loss.contiguous().mean()
             return lm_loss
 
-        (model, _), sch = slapo.build(
+        model, _ = slapo.build(
             sch,
             topology=topology,
             target="deepspeed",
@@ -130,7 +130,7 @@ def train(args):
         ds_config_dict = get_ds_config(
             batch_size, micro_batch_size, True, True, "ZeRO-3"
         )
-        (model, _), sch = slapo.build(
+        model, _ = slapo.build(
             sch,
             topology=topology,
             target="deepspeed",
--- a/examples/albert/deepspeed_hf.py
+++ b/examples/albert/deepspeed_hf.py
@@ -112,7 +112,7 @@ def train(args):
             lm_loss = lm_loss.contiguous().mean()
             return lm_loss
 
-        model, _ = slapo.build(
+        (model, _), sch = slapo.build(
             sch,
             topology=topology,
             target="deepspeed",
@@ -130,7 +130,7 @@ def train(args):
         ds_config_dict = get_ds_config(
             batch_size, micro_batch_size, True, True, "ZeRO-3"
         )
-        model, _ = slapo.build(
+        (model, _), sch = slapo.build(
             sch,
             topology=topology,
             target="deepspeed",
--- a/examples/gpt/deepspeed_hf.py
+++ b/examples/gpt/deepspeed_hf.py
@@ -150,16 +150,6 @@ def train(args):
             loss_fn=loss_fn,
             init_weights=model._init_weights,
         )
-
-        group_src_rank = _get_global_rank(sch.group, 0)
-
-        def broadcast_input(module, _inputs):
-            for inp in _inputs:
-                dist.broadcast(inp, src=group_src_rank, group=sch.group)
-            return _inputs
-
-        if model.mpu.get_pipe_parallel_rank() == 0:
-            sch["submod_0"].sync(mode="fwd_pre", sync_op_or_fn=broadcast_input)
     else:
         if batch_size is not None and micro_batch_size is None:
             micro_batch_size = batch_size // args.world_size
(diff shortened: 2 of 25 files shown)


examples/gpt/deepspeed_hf.py
```diff
@@ -125,26 +130,36 @@ def train(args):
...
+
+        group_src_rank = _get_global_rank(sch.group, 0)
+
+        def broadcast_input(module, _inputs):
+            for inp in _inputs:
+                dist.broadcast(inp, src=group_src_rank, group=sch.group)
+            return _inputs
+
+        if model.mpu.get_pipe_parallel_rank() == 0:
+            sch["submod_0"].sync(mode="fwd_pre", sync_op_or_fn=broadcast_input)
```
comaniac, line 162: This programming model is weird. Users should not add more schedules after `.build`. In other word, `.build` should not return schedule. Since users have used this primitive in schedule, we should make sure this hook is preserved when tracing the module instead of asking users to specify again.

examples/gpt/deepspeed_hf.py
```diff
@@ -165,12 +180,23 @@ def loss_fn(outputs, labels):
         model = model.to(device)
     report_memory(msg="After building model")
 
-    random_seed = 2000 + dist.get_rank()
+    if args.disable_pipeline or args.sequence_parallel:
+        random_seed = 2013 + dist.get_rank()
+    else:
+        random_seed = (
+            2013
+            + 100 * model.mpu.get_pipe_parallel_rank()
```
comaniac, line 188: Where does `.mpu` come from? Is this a DeepSpeed engine API?

examples/albert/deepspeed_hf.py
```diff
@@ -112,7 +112,7 @@ def loss_fn(outputs, labels):
             lm_loss = lm_loss.contiguous().mean()
             return lm_loss
 
-        model, _ = slapo.build(
+        (model, _), sch = slapo.build(
```
chhzh123, line 115: Why do we need to return a schedule? Users can still access `sch` even we don't return it

examples/gpt/deepspeed_hf.py
szhengac: yeah, this is a deepspeed api, since model is a deepspeed object after pipeline build

examples/albert/deepspeed_hf.py
szhengac: i think a different sch object is created in the build?

examples/gpt/schedule.py
```diff
@@ -414,9 +416,11 @@ def order_args_fn(*args, **kwargs):
 
 
 def broadcast_input(sch):
-    def broadcast_input(inputs):
+    group_src_rank = _get_global_rank(sch.group, 0)
+
+    def _broadcast_input(module, inputs):
         for inp in inputs:
```
zarzen, line 422: Some of the `inputs` are None, when `--disable_pipeline` is used, and `num_mp = 8`

examples/gpt/deepspeed_hf.py
comaniac: Please check out #27 and see if that meets the requirements of this PR. If so, please rebase this PR after merging that.

examples/gpt/schedule.py
szhengac: we will leave this one to the pr from you.

comaniac (approved):
Otherwise LGTM.

examples/gpt/schedule.py
```diff
@@ -250,7 +252,7 @@ def fwd_post_hook(_module, _input, output):
         input_mask = (_input[0] < vocab_start_index) | (_input[0] >= vocab_end_index)
         output[input_mask, :] = 0.0
         # Reduce across all the model parallel GPUs
-        dist.all_reduce(output, op=dist.ReduceOp.SUM, group=sch.group)
+        output = slapo.sharding.reduce_forward_output(output, sch.group)
```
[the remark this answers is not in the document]
szhengac: it handles the backward part, otherwise the grad will not flow through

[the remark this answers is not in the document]
comaniac: Oh I got it. This is a bit tricky tho. We need to mention in doc or tutorial that if anyone wants to have comm ops in their hooks, they have to use ours...

comaniac:
Thanks @szhengac @chhzh123 @zarzen

PR #27, opened 2023-01-30
[Schedule][replace] Transfer hooks when replacing modules

comaniac:
## Description ##
The current `.trace` did the following two steps:
1. trace the target module to be a GraphModule.
2. replace the original module with the new GraphModule.

However, the traced GraphModule won't include any hooks in the original module, so the `.sync` primitives specified before tracing, including `.trace` and `.trace_for_pipeline` are dropped. This PR fixes this issue by transferring all hooks from the original module to the new one. For example, a hook registered to broadcast inputs will be transferred 2 times during scheduling:

1. After tracing the top module, the hook is transferred from the original module to the traced module.
2. After wrapping pipeline stage to partitioned modules, the (pre-forward) hook is transferred from the original module to the first pipeline stage module (wrapped by Slapo pipeline module). 

## Checklist ##

- [x] Code is well-documented

cc @szhengac @chhzh123

### DIFF
--- a/slapo/pipeline.py
+++ b/slapo/pipeline.py
@@ -9,6 +9,7 @@ from torch.fx.passes.split_module import split_module
 
 from .logger import get_logger
 from .model_dialect import get_dialect_cls
+from .utils.common import transfer_hooks
 
 logger = get_logger()
 
@@ -424,6 +425,10 @@ def build_pipeline_model(sch, target, **kwargs):
                 stage_id_2_arg_names,
             )
         )
+
+    # Transfer hooks from the original module to the corresponding pipeline stage.
+    transfer_hooks(res_partition[0], partitioned_mod, ["fwd_pre", "bwd_post"])
+    transfer_hooks(res_partition[-1], partitioned_mod, ["fwd_post"])
 
     pipe_engine_fn = get_dialect_cls("pipeline_engine", target)
     return pipe_engine_fn(
--- a/slapo/utils/common.py
+++ b/slapo/utils/common.py
@@ -29,7 +29,7 @@ def get_hooks(mod):
     return hooks
 
 
-def transfer_hooks(old_mod, new_mod):
+def transfer_hooks(old_mod, new_mod, hook_types=None):
     """Transfer the hooks from old_mod to new_mod.
 
     Parameters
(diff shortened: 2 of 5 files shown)


slapo/pipeline.py
```diff
@@ -425,6 +426,10 @@ def build_pipeline_model(sch, target, **kwargs):
             )
         )
 
+    # Transfer hooks from the original module to the corresponding pipeline stage.
+    transfer_hooks(res_partition[0], partitioned_mod, ["fwd_pre", "bwd_post"])
```
szhengac, line 430: shouldn't we transfer all the hooks? why just specify "fwd_pre", "bwd_post" for res_partition[0]? and what bwd_post is need to be transferred?

slapo/pipeline.py
```diff
@@ -425,6 +426,10 @@ def build_pipeline_model(sch, target, **kwargs):
...
+    transfer_hooks(res_partition[-1], partitioned_mod, ["fwd_post"])
```
szhengac, line 431: similar as above

slapo/utils/common.py
```diff
@@ -0,0 +1,55 @@
...
+    mod : torch.nn.Module
+        The module.
+
+    Returns
+    -------
+    dict
+        A dictionary of hooks.
+    """
+    hooks = {"fwd_pre": [], "fwd_post": [], "bwd_post": []}
+    for hook in mod._forward_hooks.values():
```
szhengac, line 20: the order of hooks is important. are these hooks a OrderedDict?

slapo/utils/common.py
```diff
@@ -0,0 +1,55 @@
...
+        "fwd_pre": ("_forward_pre_hooks", "register_forward_pre_hook"),
+        "fwd_post": ("_forward_hooks", "register_forward_hook"),
+        "bwd_post": ("_backward_hooks", "register_backward_hook"),
+    }
+    if hook_types is None:
+        hook_types = ["fwd_pre", "fwd_post", "bwd_post"]
+
+    hook_attrs = [HOOK_TYPE_TO_ATTR[hook_type] for hook_type in hook_types]
+    for hook_attr, register_attr in hook_attrs:
+        for hook in getattr(old_mod, hook_attr).values():
```
szhengac, line 54: get_hooks is not used here?

slapo/utils/common.py
```diff
@@ -0,0 +1,55 @@
...
+        The old module.
+    new_mod : torch.nn.Module
+        The new module.
+    hook_types : Optional[List[str]]
+        The types of hooks to transfer. If None, transfer all hooks.
+    """
+    HOOK_TYPE_TO_ATTR = {
+        "fwd_pre": ("_forward_pre_hooks", "register_forward_pre_hook"),
+        "fwd_post": ("_forward_hooks", "register_forward_hook"),
+        "bwd_post": ("_backward_hooks", "register_backward_hook"),
```
szhengac, line 47: why don't we also include bwd_pre?

slapo/pipeline.py
comaniac: Note that the original hook is registered to the top module that includes pipeline modules. Now we get rid of the top module and only take its children to be pipeline stages. Thus, the fwd_pre and bwd_post hooks should be transferred to the first submodule; while fwd_post should be transferred to the last submodule.

slapo/utils/common.py
comaniac: Yes. Note that this is also the way PyTorch accesses these hooks, so if this causes any issues, then PyTorch would also have it.

slapo/utils/common.py
comaniac: No. `get_hooks` is now used in unit test.

slapo/utils/common.py
comaniac: I don't think there's a backward pre-hook in PyTorch?

szhengac: just found this: https://github.com/pytorch/pytorch/blob/b90496eef5665bc39828f6c1c522f399bcc62f3f/torch/nn/modules/module.py#L92. and it seems it does not appear in 1.12

szhengac (approved):

PR #31, opened 2023-01-31
[Benchmark] Fix ZeRO-3 step log

comaniac:
## Description ##
Fix the benchmark utility `train_with_torch` to consider micro batch when printing the log. Now it accepts optional `micro_batch_size` and only prints the loss per global batch.

## Checklist ##

- [x] Code is well-documented

cc @zarzen

### DIFF
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -92,32 +92,51 @@ def generate_pipeline_cuts(num_layers, num_pp, is_encoder_decoder=False):
 def train_with_torch(
     model,
     dataloader,
-    optimizer=None,
+    optimizer,
     preproc=None,
     postproc=None,
-    global_steps=40,
-    micro_batch_size=1,
+    steps=40,
 ):
-    """The training loop for DeepSpeedEngine and PyTorch runtime."""
-    is_deepspeed = hasattr(model, "backward")
-    if not is_deepspeed and optimizer is None:
-        raise ValueError("optimizer must be provided for PyTorch runtime")
+    """The training loop for PyTorch runtime. Note that this simple training loop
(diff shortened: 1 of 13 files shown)


examples/utils.py
```diff
@@ -108,9 +114,10 @@ def train_with_torch(
             optimizer.step()
         loss = postproc(step, loss) if postproc is not None else loss
 
-        if step % 10 == 0:
-            logger.info(f"Step {step}, LOSS: {loss.item()}", ranks=0)
+        global_step = step // micro_batch_size
```
zarzen, line 117: The `step` is the index of the micro_batch, so `step // (global_batch_size // micro_batch_size)` is the right way to count the global step. Another way is leverage the runtime: 
```python
        if model.global_steps % 10 == 0 and model.is_gradient_accumulation_boundary():
            logger.info(f"Step {model.global_steps}, LOSS: {loss.item()}", ranks=0)

        loss = None
        if model.global_steps == steps:
            break
```

comaniac:
Per offline discussion, we now have two functions to deal with different cases:
1. `train_with_deepspeed_engine`: Train with ZeRO but not pipeline. This function uses DeepSpeed model APIs `.global_steps` to print loss, so that we don't need to worry about batch size and DP size.
2. `train_with_torch`: Train with PyTorch native runtime. In this case we assume no DP, PP, and gradient accumulation, so each micro batch is just the global batch. This is currently only used by WideResNet w. TP.

In addition, this PR also reduces the number of steps in CI to reduce the CI time.

zarzen (approved):

PR #32, opened 2023-01-31
[Bugfix] Fix for sharding TP only

zarzen:
## Description ##
- When TP is enabled and PP is disabled, the last linear layer is sharded. Thus, we need `ParallelCrossEntropy` as the loss instead of the loss from HF. 
- Using TP with ZeRO-1 optimizer in DS runtime requires `mpu` as input to the `deespeed.initialize(...)` function. 
- When `--disable_pipeline` is used for disabling pipeline parallelism, we need to get the `num_mp` parameters from the `args`. 

## Known issue
- With the fix, the loss converge to about 7.x after 200 steps, which is still slower than ZeRO-1 or ZeRO-3 only. 
- For now only tested with GPT model. 

## Checklist ##

- [X] PR's title starts with a category (e.g. [Bugfix], [Model], [Tutorial], etc)
- [X] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [X] Code is well-documented

### DIFF
--- a/examples/gpt/schedule.py
+++ b/examples/gpt/schedule.py
@@ -5,8 +5,7 @@ import inspect
 
 import torch
 import torch.nn as nn
-import torch.distributed as dist
-from torch.distributed.distributed_c10d import _get_global_rank
+import torch.distributed.distributed_c10d as dist
 
 import slapo
 from slapo import init_empty_weights
@@ -416,11 +415,12 @@ def checkpoint(sch, config, path="h.N", ckpt_ratio=1.0):
 
 
 def broadcast_input(sch):
-    group_src_rank = _get_global_rank(sch.group, 0)
+    group_src_rank = dist.get_global_rank(sch.group, 0)
 
     def _broadcast_input(module, inputs):
         for inp in inputs:
-            dist.broadcast(inp, src=group_src_rank, group=sch.group)
+            if inp is not None:
+                dist.broadcast(inp, src=group_src_rank, group=sch.group)
         return inputs
 
     sch.sync(mode="fwd_pre", sync_op_or_fn=_broadcast_input)
--- a/examples/gpt/schedule.py
+++ b/examples/gpt/schedule.py
@@ -5,7 +5,7 @@ import inspect
 
 import torch
 import torch.nn as nn
-import torch.distributed.distributed_c10d as dist
+from torch.distributed import distributed_c10d as dist
 
 import slapo
 from slapo import init_empty_weights
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -12,7 +12,7 @@ _groups = []
 
 
 def get_ds_config(
-    batch_size, micro_batch_size_per_gpu, fp16=True, zero3=False, desc="", bf16=False
+    batch_size, micro_batch_size_per_gpu, fp16=True, zero_stage=0, desc="", bf16=False
 ):
     # https://github.com/microsoft/DeepSpeed/blob/ff42743/tests/unit/model_parallelism/test_configurable_parallel_pp.py#L20
     logger.info(f"fp16={fp16}, bf16={bf16}")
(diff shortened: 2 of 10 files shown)


examples/utils.py
```diff
@@ -99,18 +99,27 @@ def train_with_torch(
...
         loss = postproc(step, loss) if postproc is not None else loss
 
-        if step % 10 == 0:
-            logger.info(f"Step {step}, LOSS: {loss.item()}", ranks=0)
+        if model.global_steps % 10 == 0 and model.is_gradient_accumulation_boundary():
+            logger.info(f"Step {model.global_steps}, LOSS: {loss.item()}", ranks=0)
 
         loss = None
-        if step == steps:
+        if model.global_steps == steps:
```
comaniac, line 124: We could revert this change and rely on #31

examples/utils.py
```diff
@@ -99,18 +99,27 @@ def train_with_torch(
...
         inputs, labels = preproc(step, batch) if preproc is not None else batch
-        loss = model(*inputs, labels=labels).loss
+        # print(f"inputs from the batch: {inputs}")
+        # print(f"labels from the batch: {labels}")
+        if hasattr(model, 'loss_fn'):
+            # use customized loss_fn due to sharded linear layer
+            outputs = model(*inputs)
+            loss = model.loss_fn(outputs.logits, labels)
+        else:
+            loss = model(*inputs, labels=labels).loss
```
comaniac, line 109: This might be confusing. Maybe we could do the following:

```python
def train_with_torch(
    model, dataloader, optimizer=None, loss_fn=None, ...
):
    ...
    if loss_fn is not None:
        outputs = model(*inputs)
        loss = model.loss_fn(outputs.logits, labels)
    else:
        loss = model(*inputs, labels=labels).loss
```
Actually I could also change this in #31 so that we don't need to resolve conflicts between PRs.

slapo/model_dialect/deepspeed/engine.py
```diff
@@ -15,11 +17,17 @@ def init_ds_engine(model, **kwargs):
 
     if "config" not in kwargs:
         raise ValueError("DeepSpeed config not provided.")
+    mpu = None
+    if "topology" in kwargs:
+        mpu = kwargs["topology"]
+        if isinstance(mpu, PipeModelDataParallelTopology):
+            mpu = PipelineParallelGrid(topology=mpu)
```
comaniac, line 24:
```suggestion
    mpu = kwargs.get("topology", None)
    if mpu is not None and isinstance(mpu, PipeModelDataParallelTopology):
        mpu = PipelineParallelGrid(topology=mpu)
```

examples/gpt/schedule.py
```diff
@@ -5,8 +5,7 @@
 
 import torch
 import torch.nn as nn
-import torch.distributed as dist
-from torch.distributed.distributed_c10d import _get_global_rank
+import torch.distributed.distributed_c10d as dist
```
comaniac, line 8: Is this the same as `torch.distributed`?
Also it is recommended to use `from torch import xxx as dist` instead.

examples/gpt/schedule.py
```diff
@@ -416,11 +415,12 @@ def order_args_fn(*args, **kwargs):
 
 
 def broadcast_input(sch):
-    group_src_rank = _get_global_rank(sch.group, 0)
+    group_src_rank = dist.get_global_rank(sch.group, 0)
```
comaniac, line 418: @szhengac please double check.

comaniac (approved):
Otherwise LGTM.

examples/gpt/schedule.py
zarzen: APIs start with underscore, is not encouraged to use. APIs in the `torch.distributed` package are routed to the `torch.distributed.distributed_c10d`

examples/gpt/schedule.py
zarzen: changed the import style

comaniac:
Thanks @zarzen

comaniac:
Found that this PR breaks 3D scripts (which are not covered in CI yet). @zarzen please file a fix.

PR #35, opened 2023-02-01
[Schedule] Refactor subgraph matching

chhzh123:
## Description ##
This PR refactors the subgraph matching API and changes it to `s.find(regex_or_pattern_fn)` that is presented in the paper. Also, it adds support for more general patterns (e.g., subgraphs with multiple input arguments). A lightweight pattern language is also proposed to support fuzzy matching.

Basically, there are several use cases:
```python
# 1. Find a `call_module` node in the dataflow graph
#    we can use regex to express the module name pattern
sch.find(r"fc[0-9]")

# 2. Find a general node in the dataflow graph
#    To avoid confusion, this case is separated from the `.find()` API
#    and users can only use `.find_node()` to achieve this goal
sch.find_node(lambda node: node.name == "fc")

# 3. Find a subgraph in the dataflow graph
#    We provide different kinds of matching facilities to enable maximum flexibility
# a) Exact matching
def pattern(x: torch.Tensor):
    return F.relu(x) + x
sch.find(pattern)

# b) Use nn.functional to match a nn.module
def pattern(x: torch.Tensor):
    # `F.relu` can match `nn.ReLU` in the graph
    return F.relu(x) + x
sch.find(pattern)

# c) Use Pattern class to specify a pattern
from slapo.pattern import Pattern
class ReLUAddPattern(Pattern):
    def __init__(self):
        super().__init__()
        self.relu = nn.ReLU()

    def forward(self, x: torch.Tensor):
        return self.relu(x) + x
sch.find(ReLUAddPattern())

# d) Use pattern language
#    Here we provide `call` function as a wildcard that can be match any
#    nodes satisfying the constraints
from slapo.pattern import call
def pattern(x: torch.Tensor):
    return call("relu", x) + x
sch.find(pattern)
```

By leveraging the pattern language, we can express complex patterns in a very simple way, like the below horizontal QKV example.

```python
def pattern(x):
    x = call(r"[qkv]_proj", x)
    return x.permute(0, 2, 1, 3)
subgraph = sch.find(pattern)
```

## Checklist ##

- [x] Code is well-documented

cc @comaniac


comaniac:
Nice done. Will review tomorrow.

examples/albert/schedule.py
```diff
@@ -9,6 +9,7 @@
 
 from slapo import init_empty_weights
 from typing import Optional
+from slapo.pattern import call
```
comaniac, line 12: nit: Not sure if the term `call` would be too common. Maybe `call_pattern` or `call_node`. I don't know.

slapo/schedule.py
```diff
@@ -36,9 +38,18 @@
...
+from .pattern import Pattern, call
 
 logger = get_logger()
 
+# Wrap call as leaf
+fx.wrap(call)
+
+
+def is_lambda_function(obj):
+    return isinstance(obj, FunctionType) and obj.__name__ == "<lambda>"
```
comaniac, line 51: Put this function to the utility.

slapo/schedule.py
```diff
@@ -525,103 +537,155 @@ def find_node(self, pattern):
...
-                    res.append((name, node))
+                if isinstance(regex_or_pattern_fn, str):
+                    if node.op == "call_module" and re.match(
+                        regex_or_pattern_fn, node.target
+                    ):
+                        res.append((name, node))
+                else:
+                    # pylint: disable=else-if-used
+                    if regex_or_pattern_fn(node):
+                        res.append((name, node))
```
comaniac, line 548: Why not use `elif regex_or_pattern_fn(node)`?

slapo/schedule.py
```diff
@@ -525,103 +537,155 @@ def find_node(self, pattern):
...
-            return self.find_subgraph(node_pattern, func_pattern=func_pattern)
-        if isinstance(node_pattern, FunctionType):
-            return self.find_node(node_pattern)
-        raise RuntimeError(f"Unrecognized pattern {node_pattern}")
+        Parameters
+        ----------
+        regex_or_pattern_fn : Union[str, Callable]
+            A regular expression for specifying the target of `call_module` node, or
+            a callable function/Pattern class specifying the subgraph pattern
+        """
```
comaniac, line 683: Add  `Returns` to docstring.

slapo/schedule.py
```diff
@@ -515,8 +517,18 @@ def hook_fn(_module, _input):
...
+        """Find a node in a static dataflow graph
+
+        Parameters
+        ----------
+        regex_or_pattern_fn : Union[str, Callable]
+            If this argument is a regular expression, it will only match the `call_module` node
+            whose `target` satisfies the regex;
+            otherwise, it will try to match all the nodes satisfies the pattern function.
+            The pattern_fn should be in `lambda node: ...` format.
+        """
```
comaniac, line 530: Add  `Returns` to docstring.

slapo/schedule.py
```diff
@@ -525,103 +537,155 @@ def find_node(self, pattern):
...
+        ----------
+        regex_or_pattern_fn : Union[str, Callable]
+            A regular expression for specifying the target of `call_module` node, or
+            a callable function/Pattern class specifying the subgraph pattern
+        """
+        if isinstance(regex_or_pattern_fn, (FunctionType, Pattern)):
+            return self.find_subgraph(regex_or_pattern_fn)
+        if isinstance(regex_or_pattern_fn, str):
+            return self.find_node(regex_or_pattern_fn)
+        raise RuntimeError(f"Unrecognized pattern {regex_or_pattern_fn}")
```
comaniac, line 688: Better to show its type (`Unrecognized pattern type {type(regex_or_pattern_fn)}`) as you're using type for dispatching

tests/test_subgraph.py
```diff
@@ -0,0 +1,234 @@
...
+
+    sch = slapo.create_schedule(Model())
+
+    def pattern(x: torch.Tensor):
+        # ReLU + residual add
+        return F.relu(x) + x
+
+    subgraph = sch.find(pattern)[0]
+    assert len(subgraph) == 2
+    # pylint: disable=comparison-with-callable
```
comaniac, line 34: You can put this line on top of this file so that it works for an entire file.

comaniac (commented):
Overall LGTM. Just need some minor refactors.

examples/albert/schedule.py
chhzh123: How about just calling it `call_module`? Since we can only match module node now

slapo/schedule.py
```diff
@@ -36,9 +38,18 @@
 )
 from .tracer import trace as trace_module
 from .utils.common import transfer_hooks
+from .utils.mapping import MAPPING_FROM_FUNCTIONAL_TO_MODULE
+from .pattern import Pattern, call
 
 logger = get_logger()
 
+# Wrap call as leaf
+fx.wrap(call)
```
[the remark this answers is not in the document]
chhzh123: It is because of the limitation of torch.fx -- the leaf function wrapper can only be registered in the same module. Otherwise, the wrapper cannot work properly. See https://github.com/pytorch/pytorch/blob/v1.13.1/torch/fx/_symbolic_trace.py#L1011-L1012.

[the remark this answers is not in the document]
chhzh123: I also attach this comment in the file.

examples/albert/schedule.py
comaniac: Sounds good to me.

comaniac:
Thanks @chhzh123

PR #38, opened 2023-02-02
[Random] Random state management

comaniac:
## Description ##
- Add random state management to deal with the requirements of using the same or different random seeds in a TP group. The implementation is based on the one in Megatron-LM.
- Add an activation checkpointing with the consideration of random states.
- Add an op `DropoutWithTensorParallel` that can be replaced by users when writing a schedule.
- Add unit tests.
- Disable randomly plugin in pytest. This plugin makes our random seed setup in the test fixture useless.

Notes:
1. We now offer an API `set_random_seed` for users to call in the training script. Users have to manually call it and specify the rank of 3D parallelism.
2. All changes in this PR will have no impact if `set_random_seed` is not called in advance.
3. Fidelity testing shows the updated GPT schedule with 3D parallelism could align the loss to ZeRO-3 (with and without activation checkpointing), but flash attention has to be disabled.
4. I'll update the flash attention to the latest one and see if the problem will gone.

## Checklist ##

- [x] Code is well-documented

cc @szhengac @chhzh123

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,8 +1,9 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
+import random
 
+import numpy as np
 import pytest
-
 import torch
 from torch import distributed as dist
 
@@ -18,7 +19,6 @@ def pytest_collection_modifyitems(items):
 @pytest.fixture(scope="session")
 def init_dist(request):
     """Initialize the distributed group once in the entire test session."""
-    torch.manual_seed(9999)
     try:
         dist.init_process_group(backend="nccl")
     except Exception as err:
@@ -31,3 +31,17 @@ def init_dist(request):
             pass
 
     request.addfinalizer(destory_dist)
+
+
+@pytest.fixture(scope="function", autouse=True)
+def random_seed():
+    """Set random seed to 1) make the tests deterministic, and 2) make every
+    device generate the same weights for tensor parallelism tests.
+
+    Note that if you run pytest with "randomly" plugin enabled, this fixture
+    will have no effect. You can disable the plugin with
+    pytest -p "no:randomly" ...
+    """
+    random.seed(9999)
+    np.random.seed(9999)
+    torch.manual_seed(9999)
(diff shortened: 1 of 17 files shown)


tests/test_shard_sync_op.py
```diff
@@ -20,9 +20,6 @@
 
 def init_model_and_data(local_rank):
     model = nn.Linear(10, 10).cuda(local_rank)
-    # Make sure all devices have the same model.
-    dist.broadcast(model.weight.data, src=0)
-    dist.broadcast(model.bias.data, src=0)
```
chhzh123, line 25: Why this is no longer needed?

conftest.py
```diff
@@ -31,3 +31,17 @@ def destory_dist():
             pass
 
     request.addfinalizer(destory_dist)
+
+
+@pytest.fixture(scope="function", autouse=True)
+def random_seed():
```
chhzh123, line 37: Would `fix_random_seed` a better name?

chhzh123 (approved):
LGTM. Thanks @comaniac.

tests/test_shard_sync_op.py
comaniac: Because I disabled pytest randomly plugin.

conftest.py
comaniac: This name doesn't matter as it is an autouse fixture.

examples/gpt/megatron_hf.py
```diff
@@ -60,6 +60,10 @@ def get_model(
             delay_init=delay_init,
         )
         model, _ = slapo.build(sch, init_weights=model._init_weights)
+        # Note 1: We assume no DP and PP in this script.
+        # Note 2: This overrides Megatron random seed management, so we only use
+        #         this script for benchmarking.
+        slapo.set_random_seed(2013, None, None, sch.rank)
```
szhengac, line 66: if i understand correctly. all the DP ranks also use the same seed so the loss wouldn't be right, but we only use this script for benchmarking

szhengac:
LGTM

szhengac (approved):

PR #41, opened 2023-02-06
[Op] Add attention and bias_gelu ops

comaniac:
## Description ##
This is a prerequisite PR for adding HF GPT-2 schedule.

- Implement attention ops that use flash-attention and xformrs.
- Implement bias_gelu ops that use torchscript or torch compiler.
- Change the GPT-Neo schedule to use these ops, so that now GPT-Neo schedule doesn't depend on `epoi` anymore. Later we will update the schedules of other example models accordingly.
- [Test] Add `-rxXs` to let pytest print reasons of skipped tests.
- [Docker] Update flash-attention commit hash which improves the kernel performance by ~12%. CI image is not updated because this change doesn't impact the functionality.

## Checklist ##

- [x] Code is well-documented

@szhengac @chhzh123

### DIFF
--- a/examples/gpt/schedule.py
+++ b/examples/gpt/schedule.py
@@ -3,14 +3,12 @@
 
 import inspect
 
-import torch
 import torch.nn as nn
 from torch.distributed import distributed_c10d as dist
 
 import slapo
 from slapo import init_empty_weights
-from slapo.pattern import call_module
-from slapo.op.linear import FusedQKV
+from slapo.op import FlashSelfAttention, FlashAttentionOp, FusedMLP
 from slapo import init_empty_weights, get_cuda_rng_tracker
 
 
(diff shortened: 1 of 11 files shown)


examples/gpt/schedule.py
```diff
@@ -55,32 +53,35 @@ def replace_and_shard_attention(
...
+                if kwargs["attn_op_name"] == "native_xformers":
+                    raise RuntimeError(
+                        f"Failed to create native attention: {err}"
+                    ) from None
+
+                # Failed to use the triton kernel. This may due to unsupported
+                # GPU (< sm_75) or flash-attention is not installed. Fallback
+                # to xFormers' cutlass.
+                kwargs["attn_op_name"] = "cutlass"
+                self.module = FlashSelfAttention(**kwargs)
```
chhzh123, line 84: Can we report somewhere about what kernel we are using?

chhzh123 (approved):

examples/gpt/schedule.py
comaniac: Added.

comaniac:
Thanks @chhzh123

PR #42, opened 2023-02-07
[Example] Use .fuse() primitive when possible

chhzh123:
## Description ##
This PR fixes some issues for the example models:
* Use `.fuse()` to fuse the bias+GeLU in the MLP module. Since [TorchScript module cannot be hooked](https://github.com/pytorch/pytorch/issues/34329) and cannot properly work with DeepSpeed ZeRO-3, this feature is unset by default.
* Use subgraph matching to only optimize a small part of the Albert model instead of replacing a whole layer
* Fix Albert attention mask shape
* Remove useless functions in the schedule
* Fix an argument issue in `.replace()` and support bias+layernorm fusion

## Checklist ##

- [x] Code is well-documented


chhzh123:
I didn't change the GPT example to avoid conflicts. You can merge #41 first, and I will update other models correspondingly. @comaniac

examples/albert/model.py
```diff
@@ -27,6 +28,7 @@ def schedule_model(
     group=None,
     bcast_input=False,
     pipeline_cuts=None,
+    separate_op_fusion=False,
```
comaniac, line 31: Maybe directly using `disable_fuse_bias_gelu` would be more clear and align to `disable_flash_attn`.

examples/opt/schedule.py
```diff
@@ -187,13 +202,33 @@ def fwd_post_hook(_module, _input, output):
...
+
+            subsch = sch[prefix]
+            subsch["fc1"].decompose()
+            subsch.trace(flatten=True)
+
+            subgraphs = subsch.find(bias_gelu_pattern)
+            assert len(subgraphs) == 1
+            assert len(subgraphs[0]) == 2
+            subsch.fuse(subgraphs, compiler="TorchScript", name="FusedBiasGeLU")
+        if config.activation_function in ["gelu", "gelu_new"] and not separate_fusion:
```
comaniac, line 231: This condition a bit confusing. How about
```
if config.activation_function in ["gelu", "gelu_new"]:
    if separate_fusion:
        ...
    else:
        ...
```

tests/test_fusion.py
```diff
@@ -130,5 +132,75 @@ def pattern(x, bias):
...
+        out = sch_model(inp, resid)
+    ts_time = time.time() - start_time
+    print(f"TorchScript time: {ts_time:.4f}s")
+    start_time = time.time()
+    for _ in range(1000):
+        out_ref = mod(inp, resid)
+    vanilla_time = time.time() - start_time
+    print(f"Vanilla time: {vanilla_time:.4f}s")
+    # Performance testing
+    assert ts_time < vanilla_time
```
comaniac, line 202: This would increase the CI time and more importantly might be flaky, so I would suggest to remove.

comaniac (commented):
Otherwise LGTM

tests/test_fusion.py
chhzh123: I tested and it only took a few seconds to finish, so I guess this won't be a big deal?

examples/albert/model.py
```diff
@@ -49,6 +51,11 @@ def schedule_model(
     )
     logger.info(f"Replace {cnt} attention patterns", ranks=0)
 
+    # Operator fusion
+    if disable_fuse_bias_gelu:
```
comaniac, line 55:
```suggestion
    if not disable_fuse_bias_gelu:
```

comaniac (approved):
LGTM

chhzh123:
Please do not merge first. I'll also remove the epoi dependency.

chhzh123:
Or do you think it is better to create a separate PR? @comaniac

comaniac:
I'm fine with both, so just for your convenience.

chhzh123:
I've changed all the `epoi.attention` to `slapo.op.attention` (except for the T5 one). This PR is ready to review and merge @comaniac.

slapo/op/attention.py
```diff
@@ -436,27 +444,24 @@ def reshape_for_scores(self, x: torch.Tensor):
     def forward(
         self,
         hidden_states: Optional[tuple[torch.FloatTensor]],
-        layer_past: Optional[tuple[torch.Tensor]] = None,
         attention_mask: Optional[torch.FloatTensor] = None,
+        layer_past: Optional[tuple[torch.Tensor]] = None,
```
comaniac, line 448: Why change the argument order? In this case you also need to change gpt2/schedule.py accordingly.

chhzh123: I think it would be better to align the signature of the BertAttention. Since layer_past can be sometimes set as `None`, it may easily cause problem for module tracing and pipeline data transmission. TorchScript cannot trace this kind of module just because of the argument order.

chhzh123: I'll change the GPT2 schedule accordingly.

comaniac: Ok fair enough.

comaniac:
Thanks @chhzh123

Issue #30, opened 2023-01-31
[Tracer] PyTorch tracer does not rename SelfAttention module


chhzh123:
HuggingFace uses a very terrible variable name, `self.self`, for the `SelfAttention` module in Transformer-based models, which overwrites the Python `self` keyword and causes problems for the traced torch.fx module. (See this [line](https://github.com/huggingface/transformers/blob/v4.25.1/src/transformers/models/bert/modeling_bert.py#L394) in BERT)

Currently, we automatically rename the SelfAttention module as `self_m` in the HF tracer, but preserves the PyTorch tracer as the same. Users may encounter issues when they try to access the original SelfAttention module. We may want to unify both tracers and figure out a better way to handle these terrible naming issues.

chhzh123:
Low priority issue, just a note for future reference.

chhzh123:
It seems there is no need to rename the module at this time. So I'll close this issue.

PR #44, opened 2023-02-07
[Tracer] Remove SelfAttention renaming

### LINKED ISSUE #13: [Bug] Buffer is not maintained and redundant buffer is returned after tracing bert example
HF bert models have two buffers: `position_ids` and `token_type_ids`, so `model.state_dict` is expected to output both. However we only found `position_ids` is returned from `model.state_dict` while `token_type_ids` is missing. Also there is another redundant `mod.submod_0.submod_0.layer_0.attention.self_m.module.FusedQKV_0.fused_linear.weight` appeared in the `model.state_dict`

chhzh123:
## Description ##
This PR fixes #13 and #30 by removing the legacy code of renaming self attention module for HuggingFace models. As now we only trace the module when needed and also replace the attention module with our own module, the naming issue probably may not be a problem. We also do not refer to `self_m` in our example schedules, so I remove the renaming part from the tracer, and only prompt a warning when the tracer meets those `self` modules.

## Checklist ##

- [x] Code is well-documented

cc @szhengac @comaniac

### DIFF
--- a/slapo/tracer.py
+++ b/slapo/tracer.py
@@ -28,7 +28,6 @@ def fix_hf_module(
     for target in dir(root):
         if "_tensor_constant" in target or target in {"position_ids", "token_type_ids"}:
             submods[target] = getattr(root, target)
-    nodes_to_fix = []
     for node in root_graph.nodes:
         # Add submodule's attributes to parent module if it is used
         if node.op in {"call_module", "get_attr"} and node.target not in submods:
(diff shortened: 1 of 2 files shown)


slapo/tracer.py
```diff
@@ -39,16 +38,9 @@ def fix_hf_module(
...
-        old_target = node.target
-        new_target = old_target.replace("self", "self_m")
-        with root_graph.inserting_after(node):
-            new_node = root_graph.call_module(new_target, node.args, node.kwargs)
-            node.replace_all_uses_with(new_node)
-        root_graph.erase_node(node)
-        submods[new_target] = submods[old_target]
+            logger.warning(
+                "`self` in {root.__class__.__name__} is a Python keyword, please rename it to avoid possible conflicts."
+            )
```
comaniac, line 43: This is just a warning. What happen if we keep running?

chhzh123: I tested the traced module and it can execute normally. If we only call the SelfAttention module with the `self` prefix, it seems to work. I forget why at the first place I met the naming error, which was probably related to the previous implementation of the tracer.
```python
class Model(nn.Module):
  def __init__(self):
    self.self = ...

  def forward(self, x):
    x = self.self(x)
    ...
```

comaniac: Ok got it.

comaniac (approved):

comaniac:
Thanks @chhzh123

PR #52, opened 2023-02-11
[Schedule] Create subschedule for subgraph replacement

chhzh123:
## Description ##
This PR creates subschedule for replaced subgraph module (vertical & horizontal fusion) and attach hooks to the new module. Previously, we only create subschedule for single module replacement. Also, the `replace_function` API is changed, accepting a tuple `(str, fx.Node)` instead of a single node, in order to match the API of `replace_module`. Corresponding tests are also added to ensure correctness.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/slapo/schedule.py
+++ b/slapo/schedule.py
@@ -755,6 +755,7 @@ class SubgraphWrapper(nn.Module):
             node.replace_all_uses_with(new_node)
         self.mod.graph.erase_node(node)
 
+    # pylint: disable=too-many-branches
     def replace_module(self, new_mod, subgraphs=None, name=None):
         """Replace an entire module with a new one.
         Do NOT directly call this function, use `.replace()` instead.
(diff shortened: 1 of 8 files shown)


slapo/schedule.py
```diff
@@ -863,6 +886,29 @@ def replace_module(self, new_mod, subgraphs=None, name=None):
...
+                                old_mod, "bwd_post"
+                            ):
+                                raise RuntimeError(
+                                    f"Cannot transfer hooks from {node.target} to {name} since {node.target} has a fwd_pre/bwd_post hook"
+                                )
+                            transfer_hooks(old_mod, new_mod, ["fwd_post"])
+                        elif any(has_hook(old_mod, x) for x in hook_types):
+                            raise RuntimeError(
+                                f"Cannot transfer hooks from {node.target} to {name} since {node.target} is in the middle of the subgraph"
+                            )
```
comaniac, line 909: 1. Add a comment to explain what this loop does. The following explanation can also be used as the function docstring if this logic is extracted to a utility.
```python
# Transfer hooks of modules in the subgraph to be fused.
# For example, the fwd_pre hook of the first module in the subgraph will become
# the fwd_pre hook of the fused module. Note that if middle modules have hooks,
# we will throw errors because we cannot keep these hooks in the fused module.
```
2. Add comments to `i == 0` and `i == len(ops) - 1` to indicate they are the first and last op.

comaniac (commented):
Overall LGTM. Just some suggestions of refactoring.

comaniac (approved):
LGTM

PR #53, opened 2023-02-11
[Examples] Move examples to slapo.model_schedule

chhzh123:
## Description ##
This PR moves the schedules in the example models to `slapo.model_schedule`, so users can easily retrieve the scheduled models.

## Checklist ##

- [x] Code is well-documented


chhzh123:
The epoi is removed? It seems the test failed because of the package dependency. @comaniac

ci/task_lint.sh
```diff
@@ -15,7 +15,7 @@ bash ./scripts/lint/git-black.sh HEAD~1
 bash ./scripts/lint/git-black.sh origin/main
 
 echo "Running pylint on slapo"
-python3 -m pylint slapo --rcfile=./scripts/lint/pylintrc
+python3 -m pylint -d duplicate-code slapo --rcfile=./scripts/lint/pylintrc
```
comaniac, line 18: Since now both Slapo and tests disable this rule, we could directly disable it in pylintrc so that we don't need to specify explicitly.

slapo/model_schedule/gpt.py
```diff
@@ -1,15 +1,113 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
+"""HuggingFace GPT-Neo with model schedule."""
```
comaniac, line 3: We could also rename gpt to gpt_neo to better differentiate gpt2 (both here and examples/). This could also better connect to the name under transformers/model/gpt_neo.

comaniac (commented):
Otherwise LGTM. We could later design an API to dispatch the right "schedule_model" function based on the given model.

comaniac:
> The epoi is removed? It seems the test failed because of the package dependency. @comaniac 

It should still be there or unit tests will fail. I'm not sure why there's a lint error tho...

comaniac (approved):
LGTM

chhzh123:
I'm still changing the API. Will let you know when I'm done. @comaniac

chhzh123:
Added `apply_schedule(model, model_config, sch_config)` API.

slapo/model_schedule/api.py
```diff
@@ -0,0 +1,85 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from .registry import get_schedule_method
+from ..schedule import create_schedule
+from ..logger import get_logger
+
+
+def apply_schedule(
```
comaniac, line 9: This function specifies a schedule template and calls model specify methods, but it seems not flexible at the first glance. However it might be sufficient for now so I'll need to think more on this.

slapo/model_schedule/api.py
```diff
@@ -0,0 +1,85 @@
...
+        model_name = model_config.get("_name_or_path", None)
+    if model_name is None:
+        raise ValueError(
+            "Model name is not specified in model_config. Please use `_name_or_path` to specify the model name."
+        )
+    short_name = model_name.split("/")[-1].split("-")[0]
+    short_name = "gpt_neo" if short_name == "gpt" else short_name
+    logger = get_logger(f"{model_name}")
+    logger.info(f"Scheduling {model_name}", ranks=0)
+    logger.info(f"Schdeule config: {sch_config}", ranks=0)
```
comaniac, line 26: These two logs could be merged.

slapo/model_schedule/api.py
```diff
@@ -0,0 +1,85 @@
...
+        model.bfloat16()
+    else:
+        logger.info("Use fp32 as default model dtype", ranks=0)
+
+    group = sch_config.get("group", None)
+    sch = create_schedule(model, group=group)
+    logger.info(f"Scheduling {model_name} with TP={sch.world_size}", ranks=0)
+
+    # Tensor parallelism.
+    shard_params_fn = get_schedule_method(short_name, "shard_parameters")
```
comaniac, line 47: I didn't see the part of replacing kernels such as attention? Are they now a part of sharing? It seems not right if so.

comaniac (changes requested):
We probably need an offline discussion on Monday about the APIs.

chhzh123:
I've refactored all the code and changed the API to `apply_schedule(model, schedule_key, **schedule_config)`. Should be ready to review. @comaniac

slapo/model_schedule/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# register all model schedules in the current folder
+path = dirname(abspath(getsourcefile(lambda: 0)))
+files = [f for f in os.listdir(path) if isfile(join(path, f)) and f != "__init__.py"]
+for file in files:
+    mod = import_module(f".{file.split('.')[0]}", package="slapo.model_schedule")
+    # register the schedule method using the decorator
+    try:
+        getattr(mod, "apply_schedule")
+    except Exception:
+        pass
```
comaniac, line 21: 1. Use `hasattr` to avoid this unnecessary try-catch.

2. The schedule method has the same name as the dispatching API (apply_schedule). This seems a bit confusing, especially they have different function signature (the API has a key). Maybe using "_apply_schedule"?

slapo/model_schedule/gpt2.py
```diff
@@ -1,19 +1,133 @@
...
+        delay_init=delay_init,
+        attn_op_name=attn_op_name,
+    )
+    logger.info(
+        "Replace %d attention layers with %s op", cnt, applied_attn_op_name, ranks=0
+    )
+
+    # Replace MLP with fused kernels.
+    cnt = replace_mlp(sch[prefix], model_config, delay_init=delay_init)
+    logger.info("Replaced %d MLP layers", cnt, ranks=0)
```
comaniac, line 113: This logic is not about sharding and should not be in "shard_parameters". So do other models.

slapo/model_schedule/registry.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def get_schedule_method(model_name):
+    """Get the schedule method."""
+    if model_name not in SCHEDULE_METHODS:
+        return None
+    return SCHEDULE_METHODS[model_name]
+
+
+def register_schedule_method():
```
comaniac, line 17: Since this could also be a user API (in the case they want to register a custom schedule), I would suggest a concise name "register_schedule".

comaniac (commented):
Should be the last batch of comments.

slapo/model_schedule/__init__.py
```diff
@@ -11,11 +11,13 @@
...
+    if isfile(join(path, f)) and f not in {"__init__.py", "api.py"}
+]
 for file in files:
     mod = import_module(f".{file.split('.')[0]}", package="slapo.model_schedule")
     # register the schedule method using the decorator
-    try:
-        getattr(mod, "apply_schedule")
-    except Exception:
-        pass
+    if hasattr(mod, "apply_schedule"):
```
comaniac, line 22:
```suggestion
    if hasattr(mod, "_apply_schedule"):
```

comaniac (approved):

comaniac (changes requested):
Per offline discussion this PR still needs to fix GPT-2 schedule.

comaniac (approved):
@chhzh123 I'll leave this PR to you for the final check and merge. Thanks.

chhzh123:
I've tested. Thanks @comaniac.

PR #55, opened 2023-02-13
[Schedule] Support partial checkpointing

chhzh123:
## Description ##
This PR adds support to partial checkpointing, i.e., only checkpoint a subgraph. Users can use `.find()` primitive to obtain the required subgraph and then pass it to `.checkpoint()` primitive. A new checkpointed module will be attached to the original parent module.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/slapo/schedule.py
+++ b/slapo/schedule.py
@@ -519,6 +519,48 @@ class Schedule:
     def get_module(self, name):
         return dict(self.mod.named_modules())[name]
 
+    def construct_fx_graph(self, subgraph):
+        """Construct a new fx.Graph based on the subgraph extracted from the
+        original graph.
+
+        Parameters
+        ----------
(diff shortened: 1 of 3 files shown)


slapo/schedule.py
```diff
@@ -519,6 +519,48 @@ def hook_fn(_module, _input):
     def get_module(self, name):
         return dict(self.mod.named_modules())[name]
 
+    def construct_fx_graph(self, subgraph):
+        """Construct a new fx.Graph based on the subgraph extracted from the
+        original graph.
```
comaniac, line 524: Maybe saying this function should not be called directly? btw maybe we should rename all internal used functions such as `replace_function` by adding a prefix `_`.

comaniac (approved):
LGTM. Just need to improve comments.

PR #58, opened 2023-02-15
[Bugfix] Support tree-like subgraph matching

chhzh123:
## Description ##
For the following tree-like subgraph with multiple inputs, if we use the `pattern` to match the `+`, `-`, and `*` operators, only `+` and `*` will be returned. This is because the current implementation of subgraph matching starts from the first instruction of the original subgraph and target subgraph, and test if **users** of the instruction are the same. The process is iterative and basically is a DFS. However, this process may ignore instruction `b` since `b` is not the user of `a`.

```python
class Model(nn.Module):
    def forward(self, x, y, z):
        a = x + y
        b = y - z
        c = a * b
        return c
```
```python
def pattern(x, y, z):
    return (x + y) * (y - z)
```

To solve this problem, we can simply traverse the graph node from top to bottom following the order of the link list (i.e., visiting the **successors** in a topological order), and compare the nodes in the original and target subgraph, so that all the `+`, `-`, and `*` operators will be matched. Notice this fix still cannot support commutativity between operands, so pattern `(y - z) * (x + y)` cannot match any subgraphs. This feature may need more engineering efforts in the future.

Also, this PR adds `concrete_args` to `.replace_module()` for the case that the user-defined module has different numbers of arguments with the arguments in the original subgraph.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/slapo/schedule.py
+++ b/slapo/schedule.py
@@ -669,11 +669,9 @@ class Schedule:
                 # New matched.
                 subgraphs.append((parent_name, curr))
             matched = True
-            for curr_usr, target_usr in zip(curr.users, target.users):
-                # DFS traverse. If any subgraph is not matched, the whole graph
-                # is not matched.
+            if curr.next != curr and target.next != target:
                 matched = matched and find_match_subgraphs(
-                    curr_usr, target_usr, subgraphs
+                    curr.next, target.next, subgraphs
                 )
             return matched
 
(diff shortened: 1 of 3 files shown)


comaniac (approved):
LGTM

chhzh123:
Thanks @comaniac

PR #59, opened 2023-02-15
[Bugfix] Consolidate params with orig size

comaniac:
## Description ##
Previously we use `sch.metadata.base_params` to maintain the original parameter shape before sharding. However, the scope of schedule metadata is limited to the current schedule and won't be transferred if `.replace` is used, so `sch.metadata.base_params` won't be correct anymore in this case.

In this PR, we removed `sch.metadata.base_params` and directly set an attribute to sharded parameters to record their original shapes. In future, we should revisit the `sch.metadata` design.

## Checklist ##

- [x] Code is well-documented

cc @szhengac @chhzh123

### DIFF
--- a/slapo/schedule.py
+++ b/slapo/schedule.py
@@ -98,10 +98,6 @@ class ScheduleMetadata:
         default_factory=lambda: OrderedDict()
     )
 
-    # A mapping from parameter name to original shape
-    # Used for delay initialization
-    base_params: dict[str, tuple] = field(default_factory=lambda: DictWithValidation())
-
 
 def register_primitive(need_dist=False, finalize=False):
     """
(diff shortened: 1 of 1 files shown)


slapo/schedule.py
```diff
@@ -1366,14 +1362,24 @@ def _consolidate_and_broadcast(sch: Schedule):
...
+            cnt_shard += ret[1]
+
+        return num_params, cnt_shard
 
     if cnt_meta != 0 or cnt_materialized != 0:
-        _consolidate_and_broadcast(sch)
+        num_params, cnt_shard = _consolidate_and_broadcast(sch)
+
+    logger.info(
+        "Finished consolidating %d parameters with %d being sharded"
```
szhengac, line 1380: "parameter tensors" is better to avoid confusion

comaniac: Updated.

chhzh123 (approved):
LGTM

comaniac:
Thanks @szhengac @chhzh123

PR #68, opened 2023-02-22
[Refactor] Schedule primitives

comaniac:
## Description ##
This PR refactors the schedule primitives to make them more systematic and extensible:
* Now all primitives derived the base class `Primitive` that defines the required APIs (`name`, `apply`, and optional `init_metadata`).
* The `apply` function takes the schedule as the first argument, followed by any primitive specific arguments.
* The `init_metadata` allows each primitive to register their metadata, so the original schedule metadata only defines tie_weights metadata now. Meanwhile, we currently also put some metadata, such as the original shape before sharding, directly to nn.Parameter. I'll send a follow-up PR to better organize them.
* All primitives are registered and maintained in a table, and they will be dynamically set to a schedule when it is constructed. In other words, users can register their primitives outside of the Slapo code base.

In addition this PR also
* Move build related functions to `build.py`.
* Rename `get_output_type_after_sharding` to `postproc_sharding`, because this function not only infers the output type but also changes the attributes (e.g., `in_feature`) after sharding.

## Checklist ##

- [x] Code is well-documented

cc @chhzh123

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ We provide the following primitives for dynamic graph optimizations:
 | :--: | :-- |
 | Module replacement | `s[op].replace(new_module)` |
 | Tensor parallelism | `s[op].shard(param_name, axis)` |
-| Synchronization | `s[op].sync(mode="forward/backward/both", sync_op_or_fn, **kwargs)` |
+| Synchronization | `s[op].sync(mode="fwd_pre/fwd_post/bwd_post", sync_op_or_fn, **kwargs)` |
 | Checkpointing | `s[op].checkpoint()` |
 
 And the following primitives for static graph optimizations:
@@ -90,10 +90,24 @@ And the following primitives for static graph optimizations:
 | Module Tracing | `s.trace(leaves, flatten)` |
 | Pattern matching | `s.find(regex_or_pattern_fn)` |
 | Operator fusion | `s[op].fuse(compiler, subgraph)` |
+| Layer decomposition | `s[op].decompose()` |
 | Partial module replacement | `s[op].replace(new_module, subgraph)` |
 | Partial gradient checkpointing | `s[op].checkpoint(subgraph)` |
-| Pipeline parallelism | `s[op].pipeline_split()` |
+| Pipeline parallelism | `s[op].cut_pipeline_stage()` |
 
+You can look for all supported primitvies with the following API:
+
+```python
+import slapo
+print(slapo.list_primitives())
+```
+
+You could also check the description of each primitive on the fly:
+
+```python
+import slapo
+print(slapo.desc_primitive("shard"))
+```
 
 ### Auto-Tuning
 We also provide a light-weight interface for auto-tuning, so the developers can (1) construct a polyhedral search space using our APIs, and (2) leverage Slapo auto-tuner to automatically search for the best training configuration.
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ You could also check the description of each primitive on the fly:
 
 ```python
 import slapo
-print(slapo.desc_primitive("shard"))
+help(slapo.list_primitives(name_only=False)["shard"])
 ```
 
 ### Auto-Tuning
--- a/slapo/build.py
+++ b/slapo/build.py
@@ -1,0 +1,247 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Functions to build the model based on the schedule."""
+from __future__ import annotations
+
+import gc
+from collections.abc import Callable
+from typing import Optional, Union
+
+import torch
+from torch import distributed as dist
+from torch import nn
+
+from .framework_dialect import get_dialect_cls
+from .logger import get_logger
+from .pipeline import (
+    analyze_tie_weights,
+    build_pipeline_model,
+    generate_pipeline_partition,
+)
+from .schedule import Schedule
+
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -90,10 +90,24 @@ And the following primitives for static graph optimizations:
...
+```python
+import slapo
+print(slapo.list_primitives())
+```
+
+You could also check the description of each primitive on the fly:
+
+```python
+import slapo
+print(slapo.desc_primitive("shard"))
```
chhzh123, line 109: The name of this API is weird. Is this different from the `help` function? I'm not sure whether we need a separate API for this.

slapo/build.py
```diff
@@ -0,0 +1,247 @@
...
+    return init_engine_fn(
+        sch,
+        **kwargs,
+    )
+
+
+def build(
+    sch: Schedule,
+    target=None,
+    init_weights: Optional[Union[bool, Callable]] = True,
```
chhzh123, line 221: Should this be enabled by default? (Although I know it is already set as True in our main branch)
I guess most of the users will expect they get the same model if they do not apply schedule to the model, which means the default behavior of `model = hcl.build(sch)` should return exactly the same model as `sch.mod`. This property is not preserved since we reinitialize the weights.

slapo/primitives/checkpoint.py
```diff
@@ -0,0 +1,55 @@
...
+@register_primitive()
+class CheckpointPrimitive(Primitive):
+    """Add activation checkpointing to the entire module or a subgraph.
+
+    Parameters
+    ----------
+    subgraph : Optional[List[List[torch.fx.Node]]]
+        The subgraph to be checkpointed. If None, checkpoint the entire module.
+    order_args_fn : Optional[Callable]
+        A function to order the position and keyword arguments of the module.
```
chhzh123, line 21: Should add description about the inputs and outputs of the callable function

slapo/primitives/checkpoint.py
```diff
@@ -0,0 +1,55 @@
...
+        If None, the arguments will be ordered by the order of the position arguments,
+        followed by the key order of the keyword arguments.
+    """
+
+    @staticmethod
+    def name():
+        return "checkpoint"
```
chhzh123, line 28: Can we use `@property` instead of a `@staticmethod`?

slapo/primitives/pipeline.py
```diff
@@ -0,0 +1,63 @@
...
+        while root_sch is not None:
+            root_sch.metadata.primitives[primitive_name][parent_sch.path] = True
+            root_sch = root_sch.parent
+
+    @staticmethod
+    def init_metadata():
+        """A set of paths to the modules that includes pipeline cutting annotations.
+        Note that we use ordered set to keep the order of the modules.
+        """
+        return OrderedDict()
```
chhzh123, line 63: Why this cannot put into the `__init__` function?

chhzh123 (approved):
Otherwise LGTM

README.md
comaniac: I don't recall that we have `help` for now? I'm open with its name so let me know if you have any suggestion.

slapo/build.py
comaniac: I suppose this has no effect if the model already initialized?

slapo/primitives/checkpoint.py
comaniac: Maybe. I'll try.

README.md
chhzh123: I mean python's builtin `help` function. Users can easily use `help(sch.primitive)` to get the description of the primitive if they want.

slapo/primitives/pipeline.py
comaniac: Are you saying the `__init__` in Schedule or the `__init__` in Primitive? We cannot put this in Schedule because this is primitive specific. We cannot put it in Primitive's init neither because `Primitive` is used as a static class and will never be initialized.

slapo/build.py
chhzh123: Not really. Currently if the model has been initialized and `init_weight` is set as `True` by default, it will still reinitialize the parameters.

slapo/primitives/pipeline.py
chhzh123: I see, but we still need to preserve the data in the `Primitive` class right? If we only return a new `OrderedDict` in this case, who will keep track of its data?

comaniac: No. We preserve the data in Schedule (see schedule.py:L96). This function only provides the way to initialize the metadata along with the schedule.

README.md
comaniac: I see. Since now we register primitives on the fly, you cannot do that without initializing a schedule object; otherwise you have to `help(slapo.primitives.tensor_parallel.ShardPrimitive)`, which must be tedious. This API is more like a syntax sugar.

slapo/primitives/checkpoint.py
comaniac: Found that it's not that straightforward without adding other dependencies because we want to make it static. I'll keep the current implementation for now.

slapo/build.py
comaniac: Ok I checked the .build logic and confirmed you're right. Previously Shuai's argument was our current use cases are mainly LLMs so delay initialization is more common, but we could discuss whether there's a better way to cover all cases.

comaniac:
Thanks @chhzh123

PR #70, opened 2023-02-23
[Primitive] .fork_rng()

### LINKED ISSUE #67: [Feature] A primitive to set `get_cuda_rng_tracker()`
When sharding a model, we may need to control the RNG to maintain the correctness for convergence. For example, by default all devices in a TP group use the same RNG to make sure duplicated computations produce the same results. However, if a dropout is in a parallel region, meaning that each device in a TP group is processing a partition of activations, then each dropout should have different RNGs for better randomness. The current solution is leveraging `.replace`:

```python
class AttentionOpWithRNG(FlashAttentionOp):
    def forward(self, query_layer, key_layer, value_layer, attention_mask, p):
        with get_cuda_rng_tracker().fork():
            return super().forward(
                query_layer, key_layer, value_layer, attention_mask, p
            )

new_op = AttentionOpWithRNG(
    sub_sch["module"]["attn_op"].mod.attn_op_name,
    sub_sch["module"]["attn_op"].mod.apply_causal_mask,
    sub_sch["module"]["attn_op"].mod.scale,
)
sub_sch["module"]["attn_op"].replace(new_op)
```

Ideally we should have the something like following:

```python
# Underlying implementation is similar to ".checkpoint()"
sub_sch["module"]["attn_op"].fork_rng()
```

In terms of the primitive name, I don't have a better idea yet so you're welcome to suggest.

comaniac:
## Description ##
Close #67 

This PR adds a new primitive `.fork_rng()` that uses different random states across all devices in a TP group. This is required when the op (e.g., dropout) is working on a partitioned input in order to ensure the randomness.

Note that since now we have this primitive, we don't need `DropoutWithTensorParallel` and `AttentionOpWithRNG` ops anymore.

## Checklist ##

- [x] Code is well-documented

cc @szhengac @chhzh123


chhzh123 (approved):
LGTM

comaniac:
Thanks @chhzh123

PR #71, opened 2023-02-23
[Primitive] .annotate() and .trace_until()

comaniac:
## Description ##

This PR introduces a new primitive `.annotate()` that sets an attribute to the assigned parameter. There are mainly two use cases:
1. When sharding, we record the original shape for model consolidation.
2. Runtime such as DeepSpeed pipeline may apply certain processes based on attributes (e.g., `tensor_model_parallel` or `replicated_param`).

The annotated attribute names will be maintained in the schedule metadata of the top module. The top module schedule will not be replaced, so the metadata is guaranteed to be propagated to the model building phase. To maintain the annotations, a utility `transfer_param_tags` is also introduced.

In addition, this PR also rename `.trace_for_pipeline()` to `.trace_until()`, because this API is not only for pipelining but could also be used for other optimizations.

## Checklist ##

- [x] Code is well-documented

cc @chhzh123 @szhengac

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ We provide the following primitives for dynamic graph optimizations:
 | Synchronization | `s[op].sync(mode="fwd_pre/fwd_post/bwd_post", sync_op_or_fn, **kwargs)` |
 | Checkpointing | `s[op].checkpoint()` |
 | Fork random number generator | `s[op].fork_rng()` |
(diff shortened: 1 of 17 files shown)


chhzh123 (approved):
LGTM

comaniac:
Thanks @chhzh123

PR #74, opened 2023-02-27
[Refactor] Modulize sharding methods

### LINKED ISSUE #10: [Bug] Multiple reductions of linear bias after row sharding the weight

comaniac:
## Description ##

Close #10 

This PR refactors shard ops by modulizing necessary methods. In this way, we could easily customize the behavior of sharding a certain op/module. Accordingly, we also ported PR #23 as a demonstration.

Detail changes:
1. A new op `LinearWithSyncFunc`.
2. An API "register_shard_method" that allows each module to register a set of sharding related methods.
3. Two APIs "apply_shard_method" and "apply_sync_method" dispatches to the right method based on the module type.
4. Customize the sharding of nn.Linear and Conv1D.
5. To fix #10, in addition to performing sync op before bias addition, we also have to perform an additional allreduce on the bias gradient. Now this is achieved by annotating the bias parameters, just like layer norm.

## Checklist ##

- [x] Code is well-documented

cc @szhengac @chhzh123

### DIFF
--- a/slapo/primitives/tensor_parallel.py
+++ b/slapo/primitives/tensor_parallel.py
@@ -4,19 +4,11 @@
 # pylint: disable=arguments-differ
 
 from collections import OrderedDict
-from functools import partial
 
 from torch import nn
 
 from ..random import get_cuda_rng_tracker
-from ..sharding import (
-    all_gather_forward_output,
-    apply_shard_method,
-    reduce_backward_grad,
-    reduce_forward_output,
-    reduce_scatter_forward_output,
-    scatter_forward_output,
-)
+from ..sharding import apply_shard_method, apply_sync_method, new_or_get_tied_param
 from .base import Primitive, register_primitive
 
 
@@ -57,25 +49,7 @@ class ShardPrimitive(Primitive):
         try:
             param = sch.mod.get_parameter(tensor_name)
             new_tensor, sharded_size = _shard(tensor_name, param)
-            if param in sch.metadata.tie_weights:
-                if id(sch.metadata.tie_weights[param]) != id(param):
-                    # This parameter is tied to another parameter, and the other
-                    # parameter is already sharded. In this case we directly
-                    # register the sharded parameter to the module to keep them tied.
-                    if new_tensor.shape != sch.metadata.tie_weights[param].shape:
-                        raise RuntimeError(
-                            f"Parameter {tensor_name} in {sch.path} is tied, "
-                            "but they have different sharded shapes: "
-                            f"{new_tensor.shape} vs "
-                            f"{sch.metadata.tie_weights[param].shape}"
-                        )
-                    new_param = sch.metadata.tie_weights[param]
-                else:
-                    # The first parameter in this tie group is sharded.
-                    new_param = nn.Parameter(new_tensor)
-                    sch.metadata.tie_weights[param] = new_param
-            else:
-                new_param = nn.Parameter(new_tensor)
+            new_param = new_or_get_tied_param(sch, param, new_tensor)
             sch.mod.register_parameter(tensor_name, new_param)
 
             # Save the original size of the parameter for consolidation.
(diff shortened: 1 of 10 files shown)


slapo/primitives/tensor_parallel.py
```diff
@@ -96,75 +70,16 @@ def _shard(name, tensor):
...
-                except KeyError:
-                    raise RuntimeError(
-                        f"Output of {sch.path} has to be gathered along axis "
-                        f"{sch.metadata.primitives['shard']['gather_axis']}, but "
-                        f"{gather_axis} is requested"
-                    ) from None
-
-        out_type, out_part_axis = postproc_sharding(
-            sch.mod, tensor_name, sharded_size, axis
+        apply_shard_method("postproc", sch, tensor_name, sharded_size, axis)
```
szhengac, line 73: I feel like the new design makes what was easy a bit over complicated. 

1. What is the relationship between apply_shard_method and ShardMethod?
2. What is the api in ShardMethod such as postproc and sync used for? It seems to me these are in the old schedule's shard or sync API? And what is the relationship between this sync and old sync in schedule?

comaniac: I agree that the new design is not that straightforward, but I believe this reduces the possible ad-hoc changes in the future. For example, the previous PR #23 has to add many ad-hoc changes specifically for nn.Linear into `def shard`.

1. `apply_shard_method` determines which ShardMethod to be used. For example now we have LinearShardMethod, Conv2dShardMethod, etc. Thus, when `apply_shard_method("postproc", sch, ...)` and `sch.mod` is `nn.Linear`, we will dispatch to the sharding method of `nn.Linear`. This design lets the future developments of sharding method not change the schedule primitive directly.
2. The `postproc` is used to update the attributes (e.g., `linear.in_feature`) after sharding. This is renamed from `get_output_type_after_sharding`. The `sync` is migrated from the old sync primitive. By encapsulating this logic into ShardMethod, we are able to override it for a particular module. For example, when syncing `nn.Linear` with its input feature sharded, we have to apply all-reduce before bias addition, so we cannot simply register a hook. Instead, we replace `nn.Linear` with `LinearWithSyncFunc` and directly insert all_reduce in its forward, so that we could perform all-reduce before bias addition.

slapo/sharding/shard_ops.py
```diff
@@ -1,187 +1,611 @@
...
-    module: torch.nn.Linear
-        The linear layer to shard.
-    param_name: str
-        The name of the parameter to shard.
-    sharded_size: int
-        The size of the sharded dimension.
-    axis: int
-        The axis to shard on.
+    @staticmethod
+    def preproc(sch, param_name, sharded_size, axis):
```
chhzh123, line 252: `preproc` and `postproc` are for forward pass, then how can we handle the backward pass? Did you all reduce the bias gradient as what was done in #23?

comaniac: See (5) in description.

chhzh123 (approved):
LGTM

comaniac:
Thanks @chhzh123

PR #76, opened 2023-03-08
[Primitive][fork_rng] Do not replace module

comaniac:
## Description ##
This PR changes the way of implementing `.fork_rng()`. Now it replaces the forward function instead of an entire module. However, this approach results in incorrect trace results. Specifically, this module can still be traced, but the traced graph is based on the original forward function, which is incorrect. As a result, this PR also introduces an internal attribute `.traceable` to `nn.Module`, and we check this attribute to determine whether a Slapo customized module is traceable.

## Checklist ##

- [x] Code is well-documented

cc @chhzh123

### DIFF
--- a/slapo/op/linear.py
+++ b/slapo/op/linear.py
@@ -99,6 +99,7 @@ class LinearWithSyncFunc(nn.Linear):
         sync_fn=None,
     ):
         super().__init__(in_features, out_features, bias, device, dtype)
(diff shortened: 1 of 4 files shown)


chhzh123 (approved):
LGTM

comaniac:
Thanks @chhzh123

PR #78, opened 2023-03-09
[Primitive] Add fallback fusion

chhzh123:
## Description ##
This PR adds a fallback fusion option for `.fuse()` primitive, which directly puts the operations in the given subgraph into a `nn.Sequential` module but preserves exactly the same computation rules. It is useful for debugging and further dispatching for different backends. In this way, users even do not need to register a new compiler for Slapo, but can just replace this "fake fused" module with their efficient module using `.replace()`.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/slapo/primitives/fusion.py
+++ b/slapo/primitives/fusion.py
@@ -3,6 +3,7 @@
 """Fusion related primitives."""
 # pylint: disable=arguments-differ
 
+import types
 import torch
 from torch import nn
 
@@ -19,7 +20,7 @@ class FusePrimitive(Primitive):
     ----------
     subgraph : List[List[torch.fx.Node]]
         The subgraph to be fused.
-    compiler : str
+    compiler : Union[None, str]
         The backend compiler to be used. Currently only support "TorchScript".
     name : str
         The name of the fused module.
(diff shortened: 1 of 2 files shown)


comaniac:
This is an interesting feature. I agree that this could be useful for debugging, but I didn't get the point of replacing it with an efficient module. Shouldn't it be done with just .find and .replace?

In addition, I would suggest use torchscript as the default value instead of None, as I believe most users would expect to see speedups with .fuse.

chhzh123:
I'm imagining some backend compilers or passes to further handle these "fake fused" modules. Since the optimized fusion module may not be available ahead of time, this can be viewed as a delayed version of op fusion. TorchScript tightly couples with CPU/GPU backend and has tracer limitations, if users want another full-graph compiler to handle those fused op, using TorchScript will just make things complicated. I can think of two use cases now:
1. It actually mimics the preprocessing process of quantization in PyTorch. Specifically, [`torch.quantization.fuse_modules`](https://github.com/pytorch/pytorch/blob/master/torch/ao/nn/intrinsic/modules/fused.py) is doing the same thing as this PR does. Later on, PyTorch will take another pass to convert those fake fused modules to real fused modules on CPU.
2. If I leverage other backend compilers like HeteroCL which focuses more on operator-level optimization, it will be easier for low-level compilers to accept an encapsulated operation and then dispatch the fused op to accelerators.

This feature may not be very useful for now, but it does not break current facilities and also provides users more options to conduct graph-level optimizations.

comaniac:
So you actually meant users may want to compile the matched subgraphs in an arbitrary way, and use .replace to put the compiled module back. It makes sense to me.

comaniac:
Thanks @chhzh123

PR #80, opened 2023-03-15
[Bugfix] Shard embedding hooks

comaniac:
## Description ##
When sharding word embedding, we use forward pre-hook to mask inputs for different devices in a TP group. Accordingly, we have to mask the invalid output using the same mask in forward post-hook. Since the input mask generated in pre-hook cannot be saved, we currently re-generate the mask in post-hook. However, at this moment the input is already masked, so we cannot re-generate a mask using the same logic.

## Checklist ##

- [x] Code is well-documented

cc @chhzh123 @zarzen

### DIFF
--- a/slapo/model_schedule/gpt2.py
+++ b/slapo/model_schedule/gpt2.py
@@ -348,9 +348,8 @@ def gen_embedding_hooks(sch, vocab_size):
         return masked_input
 
     def fwd_post_hook(_module, _input, output):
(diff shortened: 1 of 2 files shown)


chhzh123 (approved):
Good catch. This bug is pretty subtle. Thanks for fixing.

comaniac:
Thanks @chhzh123

PR #82, opened 2023-03-16
[Version] Refactor version updating logic

comaniac:
## Description ##

I realized that I forgot to bump the version in `slapo/version.py` to `v0.0.2.dev0` in the last release. This results in some confusions, because when user install Slapo from source, they still see `v0.0.1.dev0`. This PR refactors the version updating logic to prevent this case from happening again in the future release.

1. Make `slapo/version.py` the only place to specify the current dev version.
2. Add a sanity check when building a wheel. Specifically, if it attempts to update the version from `v0.0.x.dev0` in version.py to `v0.0.y` where `y != x` , then errors out. This is used to make sure we bump the version in `version.py` before releasing a new stable version.
3. Align the version when installing with `setup.py`.

In short, with this PR we now have the following behaviors:
1. When installing from source (e.g., `pip3 install -e ".[dev]"`), the version shown by `pip3 show slapo` would be the formal version (e.g., `0.0.2.dev13+g44aea26`). Note that since we didn't change `slapo/version.py`, we will still see `0.0.2.dev0` if `python3 -c "import slapo; print(slapo.__version__)"`. This is intentional.
2. Before building a wheel for release, `build_wheel.sh` invokes `update_version.py`, which modifies `slapo/version.py` to be the current version. Thus, the version specified in the wheel reflects the current commit. With the sanity check added in this PR, assuming we are going to release `v0.0.3`, before adding a tag to the repo, we require the version specified in `slapo/version.py` to be `0.0.3.dev0`; otherwise `update_version.py` will error out.

## Checklist ##

- [x] Code is well-documented

cc @chhzh123

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,12 +62,17 @@ def get_version():
         "--match",
         "v[0-9]*.[0-9]*.dev[0-9]*",
     ]
-    proc = subprocess.Popen(
-        cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, cwd=ROOT_DIR
-    )
-    (out, _) = proc.communicate()
+    try:
+        proc = subprocess.Popen(
+            cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, cwd=ROOT_DIR
+        )
+        (out, _) = proc.communicate()
+        retcode = proc.returncode
+    except Exception as err:
+        out = str(err)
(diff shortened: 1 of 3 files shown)


chhzh123 (approved):
LGTM

comaniac:
Thanks @chhzh123

PR #87, opened 2023-03-26
[Verification] Add module verification support

chhzh123:
## Description ##
This PR is an initial attempt to add verification support to Slapo. For the first step, we only verify the numerical correctness by passing in random inputs and comparing the results of the original module and the scheduled one. Users can leverage the following API for verification.

```python
with slapo.verify(sch, [torch.randn(10, 20)], device=f"cuda:{local_rank}"):
    sch["linear1"].shard("weight", axis=0)
    sch["linear1"].shard("bias", axis=0)
    sch["linear1"].sync(mode="bwd_post", sync_op_or_fn="all_reduce")
    sch["linear2"].shard("weight", axis=1)
    sch["linear2"].sync(mode="fwd_post", sync_op_or_fn="all_reduce")
```

We can support verifying `.fuse()`, `.sync()`, and `.shard()` for now. If users miss synchronization at some points, the verifier will complain about the incorrect results.

## Checklist ##

- [ ] Add meta device support

Any comments and suggestions are welcome. @comaniac @zarzen


chhzh123:
We have to pass in `sch` to the verifier context in this API, since we need to store the (parent) module when entering the context. However, if we try to figure out the common parent module from the primitive sequence, then we need facilities to postpone the application of those primitives. Otherwise, the obtained parent module has been scheduled after going through all the primitives in the context. @comaniac

tests/test_verify.py
```diff
@@ -0,0 +1,110 @@
...
+            out = self.linear1(data)
+            out = self.linear2(out)
+            return out
+
+    local_rank = int(os.environ["LOCAL_RANK"])
+    torch.cuda.set_device(local_rank)
+    model = Model()
+
+    sch = slapo.create_schedule(copy.deepcopy(model))
+    with slapo.verify(sch, [torch.randn(10, 20)], device=f"cuda:{local_rank}"):
```
zarzen, line 93: Here we are passing in random inputs manually, can we automate this part base on the input shape of the linear layer?

tests/test_verify.py
```diff
@@ -0,0 +1,110 @@
...
+        sch["linear1"].shard("weight", axis=0)
+        sch["linear1"].shard("bias", axis=0)
+        sch["linear1"].sync(mode="bwd_post", sync_op_or_fn="all_reduce")
+        sch["linear2"].shard("weight", axis=1)
```
zarzen, line 97: Is it possible to support verify part of the schedule? e.g., just verify the second layer sharding.

slapo/verify.py
```diff
@@ -0,0 +1,71 @@
...
+                            break
+
+            return trace_calls
+
+        sys.settrace(trace_calls)
+        return self
+
+    def __exit__(self, *exc):
+        # Verification
+        # TODO: Support backward verification
```
zarzen, line 60: That would be great, as some of the ops are only triggered in the backward pass, e.g., all-reduce grads.

tests/test_verify.py
chhzh123: There are infinity many input tensors can be fed into the module without running into error. For example, we can pass in a 3D or 4D tensor into a linear layer. Even for shape propagation, we also need a starting shape, so we require users to explicitly pass in a random tensor here.

tests/test_verify.py
chhzh123: But how can we verify the partial results? We can for example only verify the first layer sharding, but this requires an explicit synchronization point (all_gather) after the first linear layer.

zarzen: Yes, then we should be able to insert an assertion point after all_gather, if we need/want to.

chhzh123: The point is if we automatically insert the synchronization then there will be no differences between the following two cases:
```python
with slapo.verify(sch, [torch.randn(10, 20)], device=f"cuda:{local_rank}"):
    sch["linear1"].shard("weight", axis=0)
    sch["linear1"].shard("bias", axis=0)
```
and
```python
with slapo.verify(sch, [torch.randn(10, 20)], device=f"cuda:{local_rank}"):
    sch["linear1"].shard("weight", axis=0)
    sch["linear1"].shard("bias", axis=0)
    sch["linear1"].sync(mode="bwd_post", sync_op_or_fn="all_gather")
```
which means users do not know whether they miss inserting a synchronization primitive.

chhzh123: At least the primitives inside the `verify` context should be self-contained, meaning after applying this sequence of primitives, the outputs should be identical to the original one.

zarzen: No, we don't need to insert the sync point automatically. The point is that we need the ability to break a large schedule piece by piece.

chhzh123: If you mean verify a part of the schedule, then yes, we can definitely do it, as long as that part of the schedule is self-contained, as what I shown in the second example.

comaniac: IMHO we could allow users to provide a post-process function for verification. For example, in this case, the post-processing function should have an all-gather.

slapo/verify.py
```diff
@@ -0,0 +1,71 @@
...
+                if function_name == "apply":
+                    # This part is useful only when we need to get the model from the schedule
+                    # (the schedule is not passed in as an argument)
+                    for _, value in frame.f_globals.items():
+                        cls_name = getattr(value, "__name__", None)
+                        if cls_name in {
+                            "FusePrimitive",
+                            "ShardPrimitive",
+                            "SyncPrimitive",
+                        }:
```
comaniac, line 47: Better to make this as a primitive attribute, such as a flag in primitive base class, so that we could specify it in each primitive.

tests/test_verify.py
chhzh123: > For example, in this case, the post-processing function should have an all-gather.

Then why not directly let user write a `.sync(all_gather)` primitive?

> Another idea is allowing users to custom the verify function

This probably makes things complicated for now. I think the existing interface has already placed some burden on programmers, as they are required to refactor the schedule code in order to leverage the verifier. If the verification process also needs to be customized, the schedule code will become cumbersome.

comaniac: For example when we want to verify an attention block with sequence parallel, all-gather should not be in the schedule. Just like our sharding unit tests, which we manually all-gather outputs and gradients for verification.

chhzh123:
Per offline discussion, I made the following changes: @comaniac 
1. Deep copy the scheduled model and rebuild the schedule. In the previous implementation, directly copying the model cannot work, which is because of the discarded attributes in the model. Thus, after I reattach the attributes (e.g., orig_shape) to the model and rebuild it, everything works fine.
2. Automatically broadcast the model parameters and input data.
3. Add `is_verifiable` method to the base primitive class.

slapo/verify.py
```diff
@@ -0,0 +1,106 @@
...
+from .schedule import create_schedule
+from .build import build
+from .logger import get_logger
+from .primitives import PRIMITIVES
+
+logger = get_logger()
+PRIMITIVES_NAMES = [cls.__name__ for cls in PRIMITIVES.values()]
+
+
+class verify(ContextDecorator):
```
comaniac, line 21:
```suggestion
class Verify(ContextDecorator):
```

slapo/verify.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def __enter__(self):
+        self.original_trace = sys.gettrace()
+
+        # pylint: disable=unused-argument
+        def trace_calls(frame, event, arg):
+            if event == "call":
+                code = frame.f_code
+                function_name = code.co_name
+                # local_sch = frame.f_locals.get("sch")
```
comaniac, line 39: Remove the commented code.

slapo/verify.py
```diff
@@ -0,0 +1,106 @@
...
+
+                if function_name == "apply":
+                    # This part is useful only when we need to get the model from the schedule
+                    # (the schedule is not passed in as an argument)
+                    for _, value in frame.f_globals.items():
+                        cls_name = getattr(value, "__name__", None)
+                        if cls_name in PRIMITIVES_NAMES and value.is_verifiable():
+                            logger.info("Verifying %s...", cls_name, ranks=0)
+                            break
```
comaniac, line 48: Can you just `issubclass(value, Primitive)`?

slapo/verify.py
```diff
@@ -0,0 +1,106 @@
...
+                dist.broadcast(
+                    original_state_dict[param_name], src=0, group=self.sch.group
+                )
+        # 2. Get the transformed model from the schedule
+        #    Copy it and build a new schedule to prevent the original schedule from being modified
+        copied_mod = copy.deepcopy(self.sch.mod)
+        # copy original attributes
+        for param_name, param in self.sch.mod.named_parameters():
+            if hasattr(param, "orig_shape"):
+                copied_mod.get_parameter(param_name).orig_shape = param.orig_shape
```
comaniac, line 77: This is ad-hoc to `.annotate` primitive. We should have a TODO to refactor it in the future.

slapo/verify.py
```diff
@@ -0,0 +1,106 @@
...
+        #    Notice init_weights is called before actual sharding, so we only need to
+        #    assign the original weights to the corresponding modules
+
+        def init_weights(mod, path):
+            if hasattr(mod, "weight") and mod.weight is not None:
+                mod.weight = nn.Parameter(
+                    original_state_dict[f"{path}.weight"].detach()
+                )
+            if hasattr(mod, "bias") and mod.bias is not None:
+                mod.bias = nn.Parameter(original_state_dict[f"{path}.bias"].detach())
```
comaniac, line 89: Since this is just syncing parameters, can we make it more systematic, such as the following?

```python
for name, _ in mod.named_parameters():
    setattr(mod, name, nn.Parameter(original_state_dict[f"{path}.{name}"].detach()))
```

slapo/verify.py
```diff
@@ -0,0 +1,106 @@
...
+        TODO: Support backward verification
+        """
+        # 1. Build the original model with random weights
+        named_params = self.original_sch.mod.named_parameters()
+        is_initialized = named_params.__next__()[1].device != torch.device("meta")
+        original_mod, _ = build(self.original_sch, init_weights=not is_initialized)
+        original_mod = original_mod.to(self.device)
+        #    Broadcast the original model from rank 0 to other ranks
+        original_state_dict = original_mod.state_dict()
+        if dist.is_initialized():
```
comaniac, line 66: I feel like we should refer to something like `sch.world_size`. After all, `dist.is_initialized()` doesn't mean TP is enabled.

slapo/build.py
```diff
@@ -103,7 +104,14 @@ def consolidate_model(
...
         if param_init_fn:
-            param_init_fn(sch.mod)
+            args = inspect.signature(param_init_fn).parameters
+            num_params = (
+                len(args) - 1 if args.get("self", None) is not None else len(args)
+            )
+            if num_params == 1:
+                param_init_fn(sch.mod)
+            else:
+                param_init_fn(sch.mod, sch.path)
```
comaniac, line 114: Could you elaborate this logic a bit?

comaniac (commented):
Otherwise LGTM

comaniac (approved):

PR #88, opened 2023-03-31
[Primitive] Add kwargs to .replace_all()

chhzh123:
## Description ##
This PR fixes two issues:
1. Previous implementation leverages `.named_schedule()` to get all the subschedules, which is a generator that generates the subschedules on-the-fly. However, when the user replaced module accidentally includes the original module, it will get into a infinite loop. One practical example is to replace `nn.Linear` with an inherited module or a sequential module that contains a `nn.Linear` layer. To fix it, we need to get all the subschedules first, and then replace them one by one.
2. Add `**kwargs` to `.replace_all()`, enabling users to pass additional arguments to the module constructor.

## Checklist ##

- [x] Code is well-documented

cc @comaniac

### DIFF
--- a/slapo/primitives/replace.py
+++ b/slapo/primitives/replace.py
@@ -294,6 +294,8 @@ class ReplaceAllPrimitive(Primitive):
         A target nn.Module type to be replaced.
     make_mod_fn : FunctionType
         A function that takes the path of the original module and the module itself and generate a new module.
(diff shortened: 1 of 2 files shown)


comaniac (approved):
LGTM. This is tricky lol

chhzh123:
Thanks @comaniac

PR #92, opened 2023-05-13
[Bugfix] Fix QKV matching

chhzh123:
## Description ##
This PR is an enhancement to #58. Although #58 adds support for tree-like subgraph matching, it makes the original QKV matching invalid. This is because the user of a graph node may not be the immediate successor of it, so we need to iteratively find its correct user. For the following case, `y` is not the user of `x` but is the immediate successor of `x`. When we do the pattern matching, we need to skip `y` but to match `z`.

```python
# original graph
x = a + b
y = c + d (totally independent from x)
z = x + 1
# pattern graph:
m = p + q
n = m + 1
```

In general, subgraph isomorphism is an NP-complete problem, and the current implementation is a greedy algorithm that leverages the property of sequential representation of the computation graph, but it cannot guarantee all the patterns can be matched.

## Checklist ##

- [x] Code is well-documented

cc @comaniac

### DIFF
--- a/slapo/schedule.py
+++ b/slapo/schedule.py
@@ -345,10 +345,10 @@ class Schedule:
             #  z = x + 1
             # The implication here is that the generated pattern graph should follows **the same
             # topological order** of the original graph. Otherwise, the current implementation
-            # will not be able to match the pattern graph. In general, subgraph isophorphism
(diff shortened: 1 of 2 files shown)


comaniac (approved):
LGTM

PR #93, opened 2023-05-14
[README] Update README for third-part library installation

chhzh123:
## Description ##
This PR reorganizes the README structure for benchmarking, which provides necessary instructions for users to run the test scripts in a brand-new environment. To reduce duplication, only the README file under the benchmark folder describes the installation requirements. Other README just points to this file.

It also fixes a tiny issue of Conv1DWithSyncFunc that may cause import errors when the environment does not have the transformer library.

cc @comaniac

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -20,7 +20,7 @@ pip3 install torch torchvision transformers datasets matplotlib tabulate network
 We leverage the [Flash Attention](https://arxiv.org/abs/2205.14135) kernel in [xformers](https://github.com/facebookresearch/xformers) library to accelerate the attention computation. Please follow the instructions below to install.
 
 - xformers:
-```
+```bash
 git clone https://github.com/facebookresearch/xformers.git
 cd xformers
 git checkout 48a77cc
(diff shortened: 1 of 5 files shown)


benchmark/README.md
```diff
@@ -3,10 +3,111 @@
...
 
-## Single Node with Tensor Parallelism on Megatron
+## Preparation
+
+### Install Dependencies
+
+Please use the following command to install the dependencies. By default, [PyTorch](https://pytorch.org/) 2.0 is installed for better performance.
+
+```bash
+pip3 install torch torchvision transformers datasets matplotlib tabulate networkx triton pybind11
```
comaniac, line 15: nit: This might install the wrong pytorch in terms of CUDA version.

slapo/sharding/shard_ops.py
```diff
@@ -548,7 +548,7 @@ class ShardConv1D(ShardMethod):
     and returns the output type (partial or partition) after sharding.
     """
 
-    class Conv1DWithSyncFunc(Conv1D):
+    class Conv1DWithSyncFunc(Conv1D if Conv1D is not None else nn.Module):
```
comaniac, line 551: If transformers is not available we should not run this op anyways, so I would suggest raising an exception in its constructor.

comaniac (approved):
LGTM

benchmark/README.md
chhzh123: Added clarification

slapo/sharding/shard_ops.py
chhzh123: Fixed

PR #97, opened 2023-05-29
[Tracer] Add TorchDynamo support

chhzh123:
## Description ##
This PR adds TorchDynamo as the backend tracer, providing more use cases when integrating with PyTorch 2.0.
Also, the legacy HF tracer is fixed to support tracing GPT attention modules.

## Checklist ##

- [x] Code is well-documented

cc @comaniac @zarzen @whbldhwj

### DIFF
--- a/tests/test_trace.py
+++ b/tests/test_trace.py
@@ -9,6 +9,7 @@ import torch
 from torch import fx
 
 import slapo
+from slapo.pattern import call_module
 
 
 def generate_concrete_args(model, input_names):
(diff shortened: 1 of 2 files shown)


comaniac (approved):
LGTM

tests/test_trace.py
```diff
@@ -171,5 +172,40 @@ def test_two_level_flattened_hf_bert():
...
+    )
+    subsch.trace(tracer="dynamo", concrete_args=concrete_args)
+    assert isinstance(subsch.mod, fx.GraphModule)
+
+    def pattern(x):
+        x = call_module(r"self_(query|key|value)", x)
+        x = x.view((8, 512, 16, 64))
+        return x.permute(0, 2, 1, 3)
+
+    qkv_subgraphs = subsch.find(pattern)
```
zarzen, line 202: The find API supports function as input, do we have a design doc for this? 
It is not intuitive to me, the `pattern` function seems to directly modify the computation semantics.

tests/test_trace.py
```diff
@@ -171,5 +172,40 @@ def test_two_level_flattened_hf_bert():
...
 
 
+def test_dynamo():
+    from transformers import AutoConfig, BertLMHeadModel
+
+    config = AutoConfig.from_pretrained("bert-large-uncased")
+    with slapo.init_empty_weights():
+        model = BertLMHeadModel(config)
+    sch = slapo.create_schedule(model)
+    subsch = sch[f"bert.encoder.layer.{0}.attention.self"]
```
zarzen, line 182: The same as `subsch = sch["bert.encoder.layer.0.attention.self"]`, why we need the string formatting here?

chhzh123: I'll fix this. I just copied from other test cases.

tests/test_trace.py
chhzh123: See #35. We only *find* the pattern but have not changed the computation graph here. This is used to find the Q/K/V dense layer and the consequential elementwise operators, shown in
https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt_neo/modeling_gpt_neo.py#L168-L174

chhzh123:
We need to update the CI environment to install the PyTorch 2.0 package. Otherwise, this PR cannot pass the test.

chhzh123:
Wait for #100 to be merged.

PR #99, opened 2023-05-31
[Primitive] .find() & .replace() API enhancement

chhzh123:
## Description ##
This PR refactors the `.replace()` and `.find()` primitives to let them support more cases:
1. Based on the discussion in the PyTorch [issue](https://github.com/pytorch/pytorch/issues/53534), we can wrap an fx function inside the pattern function context using `exec`. Therefore, we do not need to manually parse the source code and wrap it in a module, allowing users to write more free-form patterns, e.g. directly calling global variables, using third-party library functions. To be more specific, the following pattern is not allowed before this PR. This PR also adds a complete test case to replace the scaled dot product function in the BERT attention module.
```python
def pattern(attn_scores):
    return attn_scores / math.sqrt(  # non-builtin Python function
        config.hidden_size // config.num_attention_heads  # global variables
    )
```
2. Support replacing a horizontal or vertical pattern with another function. Previously this is not allowed, and users need to wrap the function in a module. The following coding style is now supported.
```python
def pattern(a, b):
    return a + b
def new_func(a, b):
    return a * b
subgraph = sch.find(pattern)
sch.replace(new_func, subgraph)

# Also, lambda functions enable more concise code
subgraph = sch.find(lambda a, b: a + b)
sch.replace(lambda a, b: a * b, subgraph)
```
3. Fix the insertion point of replacement. The original implementation adds the `new_func` before `mul` op, which is incorrect. This PR makes sure the insertion point of the new op is after all those ops to be replaced.
```python
def test_insertion_point():
    class Model(nn.Module):
        def forward(self, a, b, c):
            x = a + b
            y = c * 2
            z = x + y
            return z

    model = Model()
    sch = slapo.create_schedule(model)

    def pattern(a1, b1, y1):
        x1 = a1 + b1
        z1 = x1 + y1
        return z1

    subgraph = sch.find(pattern)

    def new_func(a, b, y):
        return a * b * y

    # The new_func should be inserted *after* the mul node.
    sch.replace(new_func, subgraph)
```

## Checklist ##

- [x] Code is well-documented

cc @comaniac @zarzen

### DIFF
--- a/slapo/primitives/replace.py
+++ b/slapo/primitives/replace.py
@@ -60,11 +60,72 @@ def _replace_function(sch, func, target_op):
         The string in the tuple is the name of the parent module that
         the node belongs to.
     """
-    _, node = target_op
-    with sch.mod.graph.inserting_after(node):
-        new_node = sch.mod.graph.call_function(func, node.args, node.kwargs)
+    if isinstance(target_op, list):
+        try:
+            sig = inspect.signature(func)
+        except ValueError:
+            sig = None
+        target_op = target_op[0]
+        _, first_node = target_op[0]
+        target_op = [op[1] for op in target_op]
+        subgraph_args, new_kwargs = get_required_args(sig, target_op)
+        with sch.mod.graph.inserting_before(first_node):
+            new_node = sch.mod.graph.call_function(
+                func, tuple(subgraph_args), new_kwargs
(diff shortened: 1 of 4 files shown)


comaniac (approved):
LGTM

PR #102, opened 2023-06-01
[Trace] Enhance HuggingFace tracer to support tracing at arbitrary levels

chhzh123:
## Description ##
This PR enhances the HF tracer to let it trace at arbitrary levels. It is now able to trace the GPT attention module.

We customize the tracer function in order to fit our cases. Specifically, 
1. We need to maintain the original function arguments, so these several lines of code are not needed. https://github.com/huggingface/transformers/blob/v4.28.1/src/transformers/utils/fx.py#L1078-L1103
2. We need to make sure the name of function arguments are the same as the original module, so a modification to the argument naming is needed. https://github.com/awslabs/slapo/blob/v0.0.3/slapo/tracer.py#L409-L414

## Checklist ##

- [x] Code is well-documented

cc @zarzen @comaniac

### DIFF
--- a/slapo/tracer.py
+++ b/slapo/tracer.py
@@ -594,6 +594,7 @@ def trace(model: nn.Module, **kwargs: dict[str, Any]):
                         self.orig_fns.add(orig)
 
                     try:
+                        # pylint: disable=bad-super-call
                         self.graph = super(HFTracer, self).trace(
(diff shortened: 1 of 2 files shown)


chhzh123:
I should separate the pipeline partition part into a separate PR

PR #103, opened 2023-06-02
[Pipeline] Fix incorrect argument order in the pipeline module

chhzh123:
## Description ##
This PR fixes the incorrect argument order in the pipeline module. See the following description. The previous implementation only fixes the argument order between pipeline stages, but does not handle the input mismatch.

```python
# GPT2LMHeadModel argument order
(input_ids, past_key_values, attention_mask, token_type_ids, position_ids, ...)
# Actual input arguments to the pipeline module
(input_ids, attention_mask, position_ids)
# Arguments of generated fx pipeline module 
(input_ids, position_ids, attention_mask)
```

## Checklist ##

- [x] Code is well-documented

cc @comaniac @zarzen

### DIFF
--- a/tests/test_pipeline_partition.py
+++ b/tests/test_pipeline_partition.py
@@ -10,6 +10,7 @@ from torch import nn
 from torch import fx
 
 import slapo
+from slapo.pipeline import generate_pipeline_partition
 
 
 class Pre(nn.Module):
@@ -89,7 +90,6 @@ def test_pipeline_partition():
             assert isinstance(sch["post"].mod, fx.GraphModule)
 
         sch["layers.0"].cut_pipeline_stage()
-        from slapo.pipeline import generate_pipeline_partition
 
         sch = generate_pipeline_partition(sch)
         # The traced pipeline module should include the pre and post modules.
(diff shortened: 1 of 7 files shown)


tests/test_pipeline_partition.py
```diff
@@ -129,4 +231,5 @@ def forward(self, x):
 
 
 if __name__ == "__main__":
-    pytest.main([__file__])
+    # pytest.main([__file__])
+    test_pipeline_partition()
```
comaniac, line 235: Mischange?

comaniac (approved):
Otherwise LGTM

zarzen:
What will be the args list for passing from stage to stage?

chhzh123:
> What will be the args list for passing from stage to stage?

`(hidden_state, position_ids, attention_mask)`, where `hidden_state` is the output of first stage, and `postiton_ids` and `attention_mask` are bypassed from the input arguments.

zarzen:
> 

you mean the values `(position_ids, attention_mask)` are passed to the next stage but will be ignored by the second stage?

chhzh123:
I mean `attention_mask` is required by each stage, so it should also be passed from the first stage to the second one.

I added two GPT test cases. You can check its expected output to see how the data are transferred between stages.
https://github.com/awslabs/slapo/blob/95f4b667226e5e00088bb7751c1a3fa4e5558de9/tests/test_pipeline_partition.py#L258-L277

